PR #1, opened 2023-06-28
Implement a proof-of-concept setuptools plugin

diazona:
This PR implements a simple setuptools plugin that adds a subcommand which can be invoked as

```console
$ python setup.py pyproject
```

It will print out a very basic `pyproject.toml` file incorporating some pieces of information gathered by setuptools.


sjlongland:
Had a look at the multiple-authors thing, seems it's a bit of an afterthought… apparently `flit` does handle this situation, but the original Python packaging system does not.

Ahh well, not a big worry anyway. :-)

diazona:
Yeah, we'll figure it out.

FWIW I vaguely remember from working on a different package that PyPI has some problem with listing multiple people in the author field, that's why I didn't do it in my initial PR. I forget the details though. I guess we'll deal with that when we come to it. (Hopefully at some point they will add a way to pull the newer-format author metadata from `pyproject.toml`, which does allow multiple authors)

PR #3, opened 2023-07-01
Set minimum supported Python version to 3.6

diazona:
In discussion #2, we determined that we want this package to be as widely compatible as is reasonable. Since most people are only going to use it once (or, once per package they maintain), keeping up with the latest supported Python release isn't that important, and if we require a newer version than what a package maintainer is already using, that might be enough of a barrier to discourage someone from using it. Setting the minimum version at 3.5 allows us to use type hints, so I'm implementing that here. If we find a reason that we need to restrict it to 3.6 and up, we can always change it later.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,7 +21,7 @@ packages = find:
 package_dir =
 	= src
 include_package_data = true
-python_requires = >=3.7
+python_requires = >=3.6
 install_requires =
 	tomlkit
 


diazona:
It just occurred to me: I didn't actually check that the project builds with 3.5... we should probably verify that before merging this. I can do it later.

diazona:
I went down a bit of a rabbit hole trying to figure this out, but it turns out Python 3.5 doesn't support lines like

```python
dist: setuptools.dist.Distribution = self.distribution
```

because that requires variable annotation syntax from [PEP 526](https://peps.python.org/pep-0526/), which was only implemented in 3.6. I'm sure we could work around that but I bet there are going to be other little things like that which make it an inconvenience to work with Python 3.5... @sjlongland you called it :smile: So I think it is actually better to go ahead with 3.6, at least for now, and if anyone really wants support for Python 3.5 then we can think about adding it later. I'll edit this PR accordingly.

PR #4, opened 2023-07-03
Add a simple test

diazona:
This PR sets up the infrastructure needed to test the project. Specifically, I'm adding a pytest fixture that sets up a project with a `setup.py` file and other files as specified, and using that fixture in a test that runs the pyproject conversion on a very simple project whose `setup.cfg` just defines a name and version.

Note that a full run of tests as `tox -e py310` won't pass just yet. We still need to set up the proper linting configuration.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,107 @@
+import pathlib
+import pytest
+import warnings
+from pytest_console_scripts import ScriptRunner
+from typing import Optional, Union
+
+
+class Project:
+    """
+    A Python project on which ``setup.py pyproject`` can be run. Tests should get access
+    to instances of this class by requesting the :py:func:`project()` fixture.
+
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,107 @@
...
+        :param filename:
+            A filename or ``pathlib.Path`` representing the file to write. This should
+            be a relative path, which will be interpreted relative to the project root
+            directory. If the referenced file is not inside the project root, a warning
+            will be issued.
+
+        :param content:
+            Text content to write to the file.
+        """
+        file = (self.root / filename).resolve()
```
sjlongland, line 47: Heh, I'm too addicted to `os.path`… I should really have a good close look at that `pathlib` library, that's much more intuitive than `os.path.join()` :-)

PR #7, opened 2023-07-04
Add setuptools as an explicit dependency

diazona:
I'm so used to setuptools only being required for building packages that I forgot in this case we actually depend on it directly. So I'm adding it as an explicit runtime dependency.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,6 +23,7 @@ package_dir =
 include_package_data = true
 python_requires = >=3.7
 install_requires =
+	setuptools
 	tomlkit
 
 [options.packages.find]


sjlongland:
Ahh yes, common trap… forgetting a dependency.

sjlongland:
```
remote: Resolving deltas: 100% (4/4), completed with 2 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/main.
remote: error: Changes must be made through a pull request.
```
… Ohh, tricky.  So I was trying to do the merge from the `git` CLI.  Protection didn't realise that's what I was doing, and so protectively blocked me.

That said, doing it from the CLI, one must be certain they've got the right branches and that everything is in sync, so lots to get wrong, probably better I get used to doing it through the web interface. :-)

diazona:
Ah yeah good point, I hadn't even really thought about that, I just set it up to require merge commits out of habit. But I think there is something to be said for forcing certain operations to be done through the website for traceability. One in particular that I did have in mind was forcing releases to be prepared only by CI, not manually. (but that's something you be set up later, of course)

sjlongland:
Yeah, no biggie really… historically I've liked to do the merge on the CLI as it's then digitally signed using my key so there's cryptographic proof that I did it and not someone who managed to snatch a Github session token from my browser.

In the scheme of things though, this little detail does not really matter.  The probability of this happening is quite small.  Call it security paranoia. :-)  I think the current set-up will work fine, I've just got to un-train myself from old habits.

PR #13, opened 2023-07-08
Add pre-commit hooks for remaining linters

diazona:
I'm adding pre-commit hooks for ruff and mypy, which will make it easier to not introduce code that would fail those checks. The pre-commit hooks just run the same commits we have already been running as part of CI, so there's no change to the configuration of those tools.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -48,11 +48,7 @@ class Project:
         try:
             file.relative_to(self.root)
         except ValueError:
-            warnings.warn(
-                "Writing to path {} which is not under project root {}".format(
-                    file, self.root
-                )
-            )
+            warnings.warn("Writing to path {} which is not under project root {}".format(file, self.root))
         if file.exists() and not file.is_dir():
             # This warning message is confusing if the file is a directory, so just go
             # ahead and let the write_text() call fail in that case
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,6 +10,13 @@ repos:
     # change detection by pre-commit
     #- "--exit-non-zero-on-fix"
 
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v1.4.1
+  hooks:
+  - id: mypy
+    additional_dependencies:
+    - setuptools
+
 - repo: https://github.com/psf/black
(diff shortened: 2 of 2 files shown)


tests/conftest.py
```diff
@@ -48,11 +48,7 @@ def write(self, filename: Union[pathlib.Path, str], content: str) -> None:
         try:
             file.relative_to(self.root)
         except ValueError:
-            warnings.warn(
-                "Writing to path {} which is not under project root {}".format(
-                    file, self.root
-                )
-            )
+            warnings.warn("Writing to path {} which is not under project root {}".format(file, self.root))
```
sjlongland, line 51: Heh, `black` was raising a stink about that line.  Must admit I prefer the multi-line formatting because it keeps the lines short, but we're using 120-character lines here, so it'll want to format accordingly. :-)

sjlongland (approved):

sjlongland:
Just did a rebase on `main`… see if that makes it happier.

tests/conftest.py
diazona: Haha yeah, well, in this particular case I see the appeal of the multi-line version, but in general I think that the default of 88 is painfully short and I suppose we should just put up with cases like this in the name of consistency. Though I'll definitely admit that my line-length preference of 120 is on the longer side. If you wanted to cut it down to, say, 100 or 104 or something like that, I wouldn't mind.

sjlongland: Hah, 88 painfully short, I usually use 78 because then it doesn't wrap at 80 columns.

I find with the font size (my `~/.gvimrc` has `set guifont=Terminus\ 12`, and I'm running 1366×768 right now) I use for comfortable reading, I only get 165 characters total width.  If I put two windows side-by-side to compare things, or use `vi`'s vertical-split feature, I wind up with ~80 characters visible width without wrapping anyway.

At work, things are somewhat better in that it's a large panel capable of a quite high resolution, but the practical "character space" winds up not being that much bigger, 120 chars is probably fine (been a few weeks since I've been in the office for various reasons) but it still can feel a little long for comfort.

I guess a lot comes down to what you're used to. :-)  So much of my life was lived in 80 columns text.

PR #15, opened 2023-07-08
Skip duplicate actions when pushing in this repository

diazona:
Currently when creating a pull request (at least, one not in a fork, but at the moment there are no forks anyway), we get two copies of each Github action, one for the branch push and one for the pull request. The change in this commit should skip the action running for the pull request whenever the source branch exists locally in this repository. It shouldn't affect branches coming from other people's forks, if there are forks in the future.

I copied the logic to do this from another project in which I [worked out how to do the same thing](https://github.com/pytest-dev/pytest-localserver/pull/45). The comment I've added to the workflow refers to [a blog post](https://wildwolf.name/github-actions-how-to-avoid-running-the-same-workflow-multiple-times/) I relied on when figuring out how to make this work for that other project.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,6 +62,10 @@ jobs:
         - python: pypy3.9
           platform: ubuntu-latest
     runs-on: ${{ matrix.platform }}
+    # https://wildwolf.name/github-actions-how-to-avoid-running-the-same-workflow-multiple-times/
+    if: >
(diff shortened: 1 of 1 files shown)


diazona:
Actually let's wait on this one - I want to be sure it actually works and that might involve retriggering some CI checks.

diazona:
I think it's ready to go now...

sjlongland (approved):

PR #16, opened 2023-07-08
Tweak some project settings

sjlongland:
* Add a `.gitignore` so that generated or temporary files are not accidentally committed.
* Run `pytest` in verbose mode so that on test failures, it dumps the full `diff` where mismatches were detected in the output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,19 @@
+# Ignore vi swapfiles
+*.swp
+.*.swp
+
+# Ignore Python cache directories
+__pycache__/
+
+# Ignore project coverage
+.coverage
+
+# Ignore tox working directory
+.tox/
+
+# Ignore "egg info" directories
+*.egg-info
+
+# Ignore the 'env' directory, may be used by developers for a local test
+# environment within the project workspace.
+env/
--- a/.gitignore
+++ b/.gitignore
@@ -1,19 +1,28 @@
-# Ignore vi swapfiles
-*.swp
-.*.swp
+# NB: please avoid putting environment specific rules here.  Many of these settings can be put in
+# ${HOME}/.config/git/ignore (aka ${XDG_CONFIG_HOME}/git/config, Windows users perhaps try AppData\Local?) -- then they
+# apply to all projects on your local workstation.
+#
+# Examples:
+# - vi swapfiles (*.sw?, .*.sw?)
+# - IntelliJ IDEA project (.idea/)
+# - VisualStudio Code project (.vscode/)
+#
+# See https://git-scm.com/docs/gitignore for further detail.
+
+# Distribution generated files
+/dist/
+
+# Build system generated files
+/build/
 
 # Ignore Python cache directories
 __pycache__/
 
 # Ignore project coverage
-.coverage
+/.coverage
 
 # Ignore tox working directory
-.tox/
+/.tox/
 
-# Ignore "egg info" directories
+# Ignore "egg info" directories (or files, if someone makes such an abomination)
 *.egg-info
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -0,0 +1,19 @@
+# Ignore vi swapfiles
+*.swp
+.*.swp
+
```
diazona, line 4: Not that this is _super_ important, but I've always gone by the principle recommended in the [Git documentation](https://git-scm.com/docs/gitignore) which is that editor-specific ignore rules, or anything else that will only be relevant to some developers' workflows, should go in a local ignore file like `$HOME/.config/git/ignore` (or whatever you set `core.excludesFile` to be). So I would suggest removing these lines.

```suggestion
```

.gitignore
```diff
@@ -0,0 +1,19 @@
...
+# Ignore Python cache directories
+__pycache__/
+
+# Ignore project coverage
+.coverage
+
+# Ignore tox working directory
+.tox/
```
diazona, line 12:
```suggestion
# Ignore project coverage
/.coverage

# Ignore tox working directory
/.tox/
```

because we only want to ignore these directories at the root of the repo

.gitignore
```diff
@@ -0,0 +1,19 @@
...
+
+# Ignore "egg info" directories
+*.egg-info
```
diazona, line 15: I guess if we want to limit it to directories we could do this:

```suggestion
*.egg-info/
```

although I can't imagine any circumstances in which a _file_ matching `*.egg-info` would be part of the project so it probably doesn't matter :shrug:

.gitignore
```diff
@@ -0,0 +1,19 @@
...
+
+# Ignore the 'env' directory, may be used by developers for a local test
+# environment within the project workspace.
+env/
```
diazona, line 19: This is another one that I might suggest leaving out for the same reason as the vi swap files: it's specific to each individual developer's workflow. Like, personally I use `venv` as the name when creating a local environment with `python -m venv`, or things like `dev-py36` and `dev-py310` when creating them with tox, and I'm sure anyone else who might contribute has their own naming convention for this stuff. Of course it wouldn't be hard for me to use `env/`, but I think it would be equally not hard to leave it up to each individual developer to pick their own names and add those to their local Git ignore rules, and I think that's a more scalable approach.

If anyone does commit files that shouldn't be committed, we should catch it in code review, or if there's a real concern about particular filenames we could always add a pre-commit hook to catch it as well.

diazona (commented):
Thanks for getting this going! The changes I suggested to the `.gitignore` file aren't things I will 100% insist on, but I figured we should at least discuss them before proceeding to approval.

A couple other things that might be worth including in `.gitignore` are `/build/` (used by `setup.py build` and also for generated documentation) and `/dist/` (used when generating sdists and wheels), but it doesn't matter whether we add them now or in a separate PR.

.gitignore
sjlongland: Yep, I think a lot has changed with `git` too over the years.  I've seen this done in other places, particularly with other editors (`.idea` comes to mind with those that use IntelliJ-derived development environments) or other operating systems (`.DS_Store` on MacOS X).

I learned to use `git` back in 2008 when things were a lot more primitive, so I've probably learned bad habits here. :-)  Good to learn there is a better, less viral, way of doing things.

.gitignore
sjlongland: Yep, and if someone does create a `.egg-info` _file_, we don't want it here either. :-)

diazona (approved):

PR #17, opened 2023-07-08
Implement emitting of some project dependencies

sjlongland:
This picks up `setup_requires` and `install_requires`, and puts those in the build dependencies and project dependencies TOML output.

If `setuptools` is already listed by the project, we avoid duplicating it.  Dependencies are also "normalised" in alphabetical order, with duplicates removed.

### DIFF
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -39,7 +39,7 @@ class WritePyproject(setuptools.Command):
 
         pyproject = {
             "build-system": {
-                "requires": list(sorted(setup_requirements)),
+                "requires": sorted(setup_requirements),
                 "build-backend": "setuptools.build_meta",
             }
         }
@@ -50,7 +50,7 @@ class WritePyproject(setuptools.Command):
         }
 
         # NB: ensure a consistent alphabetical ordering of dependencies
-        dependencies = list(sorted(set(dist.install_requires)))
+        dependencies = sorted(set(dist.install_requires))
         if dependencies:
             pyproject["project"]["dependencies"] = dependencies
 
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -1,6 +1,7 @@
 import setuptools
 import sys
 import tomlkit
+from pep508_parser import parser as pep508
 from typing import List, Optional, Tuple
 
 
@@ -17,9 +18,28 @@ class WritePyproject(setuptools.Command):
     def run(self):
         dist: setuptools.dist.Distribution = self.distribution
 
+        # pyproject.toml schema:
+        # https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata
+
+        # Enumerate all set-up and build requirements, ensure there are no duplicates
+        setup_requirements = set(dist.setup_requires)
+
+        # Is 'setuptools' already there?
+        seen_setuptools = False
+        for dep in setup_requirements:
+            (dep_module, _, _, _) = pep508.parse(dep)
+            if dep_module == "setuptools":
+                # Yep, here it is!
+                seen_setuptools = True
+                break
+
+        if not seen_setuptools:
+            # We will need it here
+            setup_requirements.add("setuptools")
+
         pyproject = {
             "build-system": {
-                "requires": ["setuptools"],
+                "requires": list(sorted(setup_requirements)),
                 "build-backend": "setuptools.build_meta",
             }
         }
@@ -28,5 +48,10 @@ class WritePyproject(setuptools.Command):
             "name": dist.get_name(),
             "version": dist.get_version(),  # TODO try to reverse-engineer dynamic version
         }
+
+        # NB: ensure a consistent alphabetical ordering of dependencies
+        dependencies = list(sorted(set(dist.install_requires)))
+        if dependencies:
+            pyproject["project"]["dependencies"] = dependencies
 
         tomlkit.dump(pyproject, sys.stdout)
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -25,15 +25,9 @@ class WritePyproject(setuptools.Command):
         setup_requirements = set(dist.setup_requires)
 
         # Is 'setuptools' already there?
-        seen_setuptools = False
-        for dep in setup_requirements:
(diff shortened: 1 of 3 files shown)


diazona (changes requested):
Nice to see the main functionality moving forward :smile:

Don't take this to imply that any of what you did here is somehow "wrong", but I have some thoughts on how the tests are going to work. When I was writing that first test, it occurred to me that it's kind of inefficient to have the plugin generate a data structure and write it out into TOML, only to parse the TOML back into a data structure again. So I was thinking of refactoring the plugin to separate the part that actually generates the `pyproject` data structure from the part that writes that out into a TOML file, and that way most of our tests could be unit tests on the former part, so they wouldn't have to go through the full complexity of rendering and then re-parsing the TOML. We could just keep a couple of tests that write out the TOML file to make sure that the file-writing part doesn't fail.

Though, now that I think about it, even if you're on board with that, it is very possible that the best way to move forward is to just merge this and then I can refactor the tests in a separate PR to show you what I mean. What do you think?

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -17,9 +18,28 @@ def finalize_options(self):
...
+                break
+
+        if not seen_setuptools:
+            # We will need it here
+            setup_requirements.add("setuptools")
+
         pyproject = {
             "build-system": {
-                "requires": ["setuptools"],
+                "requires": list(sorted(setup_requirements)),
```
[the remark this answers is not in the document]
sjlongland: Ahh okay, I thought it returned a generator… I'll fix that.

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -17,9 +18,28 @@ def finalize_options(self):
...
+        setup_requirements = set(dist.setup_requires)
+
+        # Is 'setuptools' already there?
+        seen_setuptools = False
+        for dep in setup_requirements:
+            (dep_module, _, _, _) = pep508.parse(dep)
+            if dep_module == "setuptools":
+                # Yep, here it is!
+                seen_setuptools = True
+                break
```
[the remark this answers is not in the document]
sjlongland: Yep, it's more succinct… but not quite there:

```
>       assert result_parsed == reference_parsed
E       AssertionError: assert {'build-syste...on': '0.0.1'}} == {'build-syste...on': '0.0.1'}}
E         Omitting 1 identical items, use -vv to show
E         Differing items:
E         {'build-system': {'requires': ['pytest-black<99.88.77', 'pytest>=6', 'setuptools', 'setuptools>=34.56', 'sphinx'], 'build-backend': 'setuptools.build_meta'}} != {'build-system': {'requires': ['pytest-black<99.88.77', 'pytest>=6', 'setuptools>=34.56', 'sphinx'], 'build-backend': 'setuptools.build_meta'}}
E         Use -v to get more diff

tests/test_setup_command.py:18: AssertionError
```

Let me try some brackets…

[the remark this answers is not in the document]
sjlongland: Duh, forgot the `[0]`… that'd be why.  Morning cup of tea not yet kicked in. :-)

diazona (approved):

tests/test_setup_command.py
```diff
@@ -1,4 +1,27 @@
...
+    assert result.stdout.startswith(prefix)
+
+    # Parse the reference in case of formatting differences
+    reference_parsed = tomlkit.parse(reference)
+
+    # Parse the resultant text:
+    result_parsed = tomlkit.parse(result.stdout[len(prefix) :])
+
+    # Assert equivalence
+    assert result_parsed == reference_parsed
```
diazona, line 18: I missed this before, but the comments on lines 14 and 17 seem kind of redundant (somewhat along the lines of the proverbial `i++; // increment i`), so I'd suggest removing them, but this is not a big deal and not something I'll insist on. Let me know if you really want to keep these.

```suggestion
    result_parsed = tomlkit.parse(result.stdout[len(prefix) :])

    assert result_parsed == reference_parsed
```

tests/test_setup_command.py
```diff
@@ -16,5 +39,134 @@ def test_name_and_version(project) -> None:
...
+dependencies = ["dependency1", "dependency2>=1.23", "dependency3<4.56"]
+"""
+    project.setup_cfg(setup_cfg)
+    project.setup_py()
+    result = project.run()
+    check_result(result, pyproject_toml)
+
+
+# setup_requires tests: we have to use real dependencies that actually are
+# installed, otherwise the tests will fail.
```
diazona, line 80: I hope at some point we can figure out a way around this, but nothing to do about it for now.

diazona (approved):
Just one last suggestion about the possibly-redundant comments, but I'm approving this regardless because that's not worth blocking the merge. I leave it up to you @sjlongland to decide whether to remove those comments or not, and then once you do, feel free to merge it yourself. (Or you can re-request review and I'll merge it if you feel better doing it that way.)

tests/test_setup_command.py
```diff
@@ -1,4 +1,27 @@
...
+
+def check_result(result, reference, prefix="running pyproject\n"):
+    """
+    Check the result succeeded, and matches the expected output.
+    """
+    assert result.returncode == 0
+    assert result.stdout.startswith(prefix)
+
+    # Parse the reference in case of formatting differences
+    reference_parsed = tomlkit.parse(reference)
```
diazona, line 12: Oh by the way: tomlkit's documentation says that it's a style-preserving TOML library, meant to preserve indentation, comments, etc. That's part of why I chose to use it over `tomli`, so we can get output with predictable formatting. So I think we shouldn't have to worry that much about formatting differences - or in other words, if the formatting does come out different, it means that part of the system is actually not doing what it's supposed to do, it wouldn't just be a random fluke. But we can always worry about that later.

tests/test_setup_command.py
sjlongland: Yep, I was sort-of on auto-pilot writing those comments.  They've been removed.

diazona (approved):

diazona:
Hmm I'm using the app and it's not showing me what is blocking the merge now... Assuming my approval went through, feel free to merge this yourself @sjlongland, and if it didn't then I'll figure this out when I get back to my computer.

sjlongland:
Yeah, I think Github got its knickers in a twist over the comments that I removed, it didn't notice the suggestion was basically implemented and that the comment was now resolved. :-)

PR #21, opened 2023-07-10
Add complete TOML type information

diazona:
This commit creates type aliases for the types involved in a structure that can be rendered to TOML. This will help us use mypy to ensure that we're not adding anything to the returned data structure that couldn't be written out in TOML format.

In order to help convince mypy that this doesn't introduce a type violation, I had to add the setuptools type stubs as a testing dependency.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,6 +26,7 @@ install_requires =
 	pep508-parser
 	setuptools
 	tomlkit
+	typing-extensions >=4, <5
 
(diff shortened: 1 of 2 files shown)


diazona:
After #17 is merged, I'll rebase this and then mark it as ready.

diazona:
This one should be good to go

sjlongland (approved):

Issue #20, opened 2023-07-09
Rewrite release logic


diazona:
I have a few changes in mind for the release logic:

- Make it so that a release can only be uploaded to PyPI from Github (or more precisely, remove the logic that automates publishing a release with one command in order to make it _difficult_ to upload a release without using Github). This is for traceability: it should be possible for anyone reviewing the project to see exactly how each release was generated and uploaded, and so that we have the "audit trail" of a Github tag and associated release notes for every release that makes it to PyPI.
- Use [trusted publishing](https://docs.pypi.org/trusted-publishers/) instead of a token for authenticating to PyPI.
- Upload any release to Test PyPI first to ensure that the upload works.

diazona:
Everything I listed in the initial description has been done so I think I can close this now.

PR #23, opened 2023-07-12
Test additional Python versions and dependency versions

diazona:
This PR introduces tests for Python 3.6 and PyPy 3.7 and 3.8, as well as tests with the minimum declared versions of dependencies.

This depends on #22.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,6 +27,8 @@ jobs:
         # probably work on Mac and Windows too. But if an OS-specific bug does
         # slip through, we should catch it in pre-release testing.
         - ubuntu-latest
+        force-minimum-dependencies:
+        - false
         exclude:
         # Python 3.6 does not run on ubuntu-latest
         - python: "3.6"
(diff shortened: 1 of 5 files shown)


diazona:
Apparently this fails tests on PyPy 3.7 so we'll have to figure out some way to work around that.

sjlongland:
Yeah, I should install Pypy and see how well some of my other stuff works.  While not as commonly used, from what I understand it's a very fast implementation of Python.

diazona:
That's what I hear as well. I've never really found a need to use it myself, but I can see how it would be beneficial with some particularly complex scripts or long-running operations. Some people used it for a few things at my prior-to-current job, and for the most part it was a drop-in replacement for CPython, but there are definitely some edge cases.

diazona:
Based on the latest test results I think this is going to be blocked by #20, so I'll work on that soon.

sjlongland (approved):
Looks good… I think constraining the OSes we test against until we go to release is fine, can't be cheap for Github to license runners with these OSes.

PR #24, opened 2023-07-12
Add more pre-commit hooks

diazona:
Pretty simple, this PR adds some basic hooks from https://github.com/pre-commit/pre-commit-hooks to do things like checking whitespace and syntax.


sjlongland (approved):
Definitely learning some validation tricks with this project. :-)

PR #25, opened 2023-07-15
Refactor testing to use the data structure directly

diazona:
In this PR I'm splitting out most of the functionality of the plugin's `run()` method into a new method, `_generate()`, which returns the data structure corresponding to the contents of the `pyproject.toml` file but without actually writing it out as TOML. I've also modified the test infrastructure to give access to the return value of this method directly, through the new `Project.generate()` wrapper, so we can run tests without going through an unnecessary (ish?) round of writing the structure to TOML and then reading it back in.

I also copied all the existing tests but changed them to use this new `generate()` method. This introduces a bunch of duplicate tests. I think we can safely get rid of many of the original tests and just keep a few to make sure the whole process works, but most of our testing from now on can probably be done by using `generate()`. As I create this PR I'm not yet sure which/how many of the original tests it makes sense to remove, but we'll figure that out. @sjlongland I'd appreciate your thoughts on this.

### DIFF
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -54,7 +54,11 @@ class WritePyproject(setuptools.Command):
     def finalize_options(self):
         pass
 
-    def run(self):
+    def _generate(self) -> Pyproject:
+        """
+        Create the raw data structure containing the information from
+        a pyproject.toml file.
+        """
         dist: setuptools.dist.Distribution = self.distribution
 
         # pyproject.toml schema:
@@ -87,7 +91,14 @@ class WritePyproject(setuptools.Command):
         if dependencies:
             pyproject["project"]["dependencies"] = dependencies
 
-        tomlkit.dump(pyproject, sys.stdout)
+        return pyproject
+
+    def run(self):
+        """
+        Write out the contents of a pyproject.toml file containing information
+        ingested from ``setup.py`` and/or ``setup.cfg``.
+        """
+        tomlkit.dump(self._generate(), sys.stdout)
 
 
 __all__ = ["WritePyproject"]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,11 @@
+import distutils.core
+import distutils.dist
 import pathlib
 import pytest
+import setuptools
 import warnings
 from pytest_console_scripts import ScriptRunner
+from setuptools_pyproject_migration import Pyproject, WritePyproject
 from typing import Optional, Union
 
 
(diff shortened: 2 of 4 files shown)


diazona:
I still have a bit of work to do to fix mypy's complaints about assignments to `pyproject`... I'm working on that. But other than the type annotations this should be ready.

sjlongland:
Yeah, that would remove a source of potential test failures.  It'd still be worth keeping a test that ensures that valid TOML is emitted regardless, but we'd then be able to make it just one single test and not multiple like we have now.

diazona:
Yay, I finally managed to satisfy mypy :grin: It was a bit tricky in this case.

@sjlongland Any thoughts about which of the original tests in `test_setup_command.py` we should keep to ensure that valid TOML is emitted? Personally, I'm thinking that it makes sense to keep one very basic test that emits a minimal `pyproject.toml` (like, just the project name or something), and one test that is as comprehensive as possible, emitting all the fields we have defined - of course we should keep updating that test as we add support for new fields. And honestly I'm not even sure of what sort of problem the basic test would detect; it's just my instinct that it might be useful to have as an edge case. :shrug: Let me know if you think there are other cases that would be useful to keep, and I'll push a commit that removes the ones we don't need. Or you could push that commit yourself, if you prefer.

sjlongland:
> Yay, I finally managed to satisfy mypy grin It was a bit tricky in this case.
> 
> @sjlongland Any thoughts about which of the original tests in `test_setup_command.py` we should keep to ensure that valid TOML is emitted?
> Personally, I'm thinking that it makes sense to keep one very basic test that emits a minimal `pyproject.toml` (like, just the project name or something), and one test that is as comprehensive as possible, emitting all the fields we have defined - of course we should keep updating that test as we add support for new fields.

`test_name_and_version` seems the simplest… the others are testing the logic in behind the generation of the data that is transformed to TOML, so would be good candidates to be re-worked to call the underlying `generate()` function.

I don't think there's one that emits _everything_ yet.

> And honestly I'm not even sure of what sort of problem the basic test would detect; it's just my instinct that it might be useful to have as an edge case. shrug Let me know if you think there are other cases that would be useful to keep, and I'll push a commit that removes the ones we don't need. Or you could push that commit yourself, if you prefer.

True, and if we keep the serialisation routine as dumb as possible, we're realistically not testing our own code, we're testing `tomlkit`, which should have its own tests.

Another approach is the call to `tomlkit` is its own method, which in unit tests, we stub and is itself excluded from test coverage (as we keep it "trivial").  That way, we just test what we feed _into_ `tomlkit`, and we assert whatever we return from the stub, is processed correctly by the calling function.

It comes down to how much you trust the upstream project to be consistent in their output. :-)

diazona:
> test_name_and_version seems the simplest… the others are testing the logic in behind the generation of the data that is transformed to TOML, so would be good candidates to be re-worked to call the underlying generate() function.

Makes sense. I'll push a commit that removes the others.

> I don't think there's one that emits everything yet.

Yeah, I mean we haven't implemented support for most of the fields but I think there should be one test that emits all the fields we do have support for, and as we add support for new fields we just keep expanding the scope of that test. Sorry if I wasn't clear about that.

> True, and if we keep the serialisation routine as dumb as possible, we're realistically not testing our own code, we're testing tomlkit, which should have its own tests.

My thoughts as well :+1: 

> Another approach is the call to tomlkit is its own method, which in unit tests, we stub and is itself excluded from test coverage (as we keep it "trivial"). That way, we just test what we feed into tomlkit, and we assert whatever we return from the stub, is processed correctly by the calling function.
>
> It comes down to how much you trust the upstream project to be consistent in their output. :-)

Hmm, interesting idea. I hadn't thought of that. I do think it makes the test code a little more complicated to follow, though probably not enough to be a big deal. FWIW I do trust `tomlkit`'s output to be stable over the long term; or if it's not, we can probably adjust for that (if nothing else by parsing the output as you implemented before), but bypassing `tomlkit` entirely does save us that particular bit of effort.

sjlongland (approved):

PR #45, opened 2023-08-05
Add support for author and maintainer fields

### LINKED ISSUE #28: Support authors and maintainers
This issue entails adding support for the `authors` and `maintainers` fields in `pyproject.toml` when they are defined statically in the setuptools configuration.

diazona:
This pull request adds support for the author and maintainer fields in the plugin, and adds a bunch of tests to verify that it's working properly.

I made some arbitrary (but sensible) choices about how to handle unusual cases, like names without email addresses or vice versa, or extra commas - basically it just substitutes in an empty string any time something would be missing. I don't know if that aligns with how other tools work, but if we need to change that behavior later, it'll be easy to do so.

Closes #28

### DIFF
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -1,3 +1,4 @@
+import itertools
 import setuptools
 import sys
 import tomlkit
@@ -54,6 +55,51 @@ class WritePyproject(setuptools.Command):
     def finalize_options(self):
         pass
 
+    @staticmethod
+    def _strip_and_canonicalize(s: str) -> str:
+        """
+        Strip whitespace from around a string, but replace the sentinel value
+        ``"UNKNOWN"`` (used by setuptools<62.2) with an empty string.
+
+        >>> WritePyproject._strip_and_canonicalize("        ooh, you lucky bastard")
+        'ooh, you lucky bastard'
+        >>> WritePyproject._strip_and_canonicalize("UNKNOWN")
+        ''
+        >>> WritePyproject._strip_and_canonicalize("")
+        ''
+        """
+        s = s.strip()
+        if s == "UNKNOWN":
+            return ""
+        else:
+            return s
+
+    @staticmethod
+    def _transform_contributors(name_string: Optional[str], email_string: Optional[str]) -> List[Contributor]:
+        """
+        Transform the name and email strings that setuptools uses to specify
+        contributors (either authors or maintainers) into a list of dicts of
+        the form that should be written into ``pyproject.toml``.
+
+        >>> WritePyproject._transform_contributors("John Cleese", "john@python.example.com")
+        [{'name': 'John Cleese', 'email': 'john@python.example.com'}]
+
+        Missing entries will be replaced with the empty string.
+
+        >>> WritePyproject._transform_contributors("John Cleese, Graham Chapman", "john@python.example.com")
+        [{'name': 'John Cleese', 'email': 'john@python.example.com'}, {'name': 'Graham Chapman', 'email': ''}]
+
+        :param: name_string  A string giving a comma-separated list of contributor
+                             names.
+        :param: email_string A string giving a comma-separated list of contributor
+                             email addresses which correspond to the names.
+        :returns:            A list of dicts containing corresponding names and
+                             email addresses parsed from the strings.
+        """
+        names = map(WritePyproject._strip_and_canonicalize, (name_string or "").split(","))
+        emails = map(WritePyproject._strip_and_canonicalize, (email_string or "").split(","))
+        return [{"name": n, "email": e} for n, e in itertools.zip_longest(names, emails, fillvalue="") if n or e]
+
     def _generate(self) -> Pyproject:
         """
         Create the raw data structure containing the information from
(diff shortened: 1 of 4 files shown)


diazona:
I'll have to come back to this and figure out why it's breaking on 3.6

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -54,6 +55,12 @@ def initialize_options(self):
     def finalize_options(self):
         pass
 
+    @staticmethod
+    def _transform_contributors(name_string: Optional[str], email_string: Optional[str]) -> List[Contributor]:
+        names = map(str.strip, (name_string or "").split(","))
+        emails = map(str.strip, (email_string or "").split(","))
+        return [{"name": n, "email": e} for n, e in itertools.zip_longest(names, emails, fillvalue="") if n or e]
```
sjlongland, line 62: Learning so many little tricks with this project. :-)

sjlongland (approved):

sjlongland:
Ahh of course, I'm jumping the gun reviewing this with Python 3.6 being recalcitrant … but I can give it another quick look afterwards.

sjlongland:
Ahh… spotted what is wrong… change in behaviour between Python 3.10 and later.

Dummy project for testing:
```
RC=0 stuartl@rikishi /tmp/mockpkg $ tail -v -n +0 *
==> setup.py <==
from setuptools import setup

setup(name="mockpkg", version="1.2.3")
```

In Python 3.6:
```
$ docker run --rm -ti -w /tmp/test -v $PWD:/tmp/test:ro python:3.6-alpine ash
/tmp/test # python3 setup.py --author
UNKNOWN
/tmp/test # python3 setup.py --author-email
UNKNOWN
```

Python 3.9:
```
$ docker run --rm -ti -w /tmp/test -v $PWD:/tmp/test:ro python:3.9-alpine ash
/tmp/test # python3 setup.py --author
UNKNOWN
/tmp/test # python3 setup.py --author-email
UNKNOWN
```

Python 3.10:
```
$ docker run --rm -ti -w /tmp/test -v $PWD:/tmp/test:ro python:3.10-alpine ash
/tmp/test # python3 setup.py --author
None
/tmp/test # python3 setup.py --author-email
None
```

You might need to check for the name/email being `"UNKNOWN"` as well as it being falsey.

src/setuptools_pyproject_migration/__init__.py
sjlongland, line 62: So the `if n or e` is where we come stuck in Python ≤ 3.9.

A quick way I can think of fixing this (may not be the most elegant):

`any([s and (s != "UNKNOWN") for s in (n, e)])`

This seems to have the logic we want:

```
RC=0 stuartl@rikishi /tmp/mockpkg $ ipython3
Python 3.11.4 (main, Jul 14 2023, 09:30:02) [GCC 12.2.1 20230428]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.12.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: defined = lambda n, e : any([s and (s != "UNKNOWN") for s in (n, e)])

In [2]: defined("foo", "bar@example.com")
Out[2]: True

In [3]: defined("foo", None)
Out[3]: True

In [4]: defined(None, "bar@example.com")
Out[4]: True

In [5]: defined(None, None)
Out[5]: False

In [6]: defined("UNKNOWN", None)
Out[6]: False

In [7]: defined(None, "UNKNOWN")
Out[7]: False

In [8]: defined("UNKNOWN", "UNKNOWN")
Out[8]: False
```

sjlongland (changes requested):

src/setuptools_pyproject_migration/__init__.py
diazona: Oh that's sneaky :laughing: Thanks for chasing this down!

The thing that comes to mind with that approach is that it will still emit the record if the name is present but the email address isn't, or vice versa, and then we could get something like `{"name": "foo", "email": "UNKNOWN"}`. And I don't think it makes sense to use `"UNKNOWN"` as a value in any context unless we're consistently using it everywhere, which we're not. (Maybe we should? :shrug: I think not, unless we find a good reason to.) With that in mind I would rather capture the `"UNKNOWN"`s in the name string and email string and replace them with `None`. Let me implement that, I'll push another commit and let you know.

diazona: Pushed f80e21cb6a6618153f0af38d0721a21c38bccd78 with a fix, which I will need to squash before merging, but feel free to take a look and let me know what you think.

diazona:
I went dependency-hunting and found that the change in behavior with respect to `UNKNOWN` comes from [PR 138 in `distutils`](https://github.com/pypa/distutils/pull/138), which was [incorporated into `setuptools` in PR 3311](https://github.com/pypa/setuptools/pull/3311), which landed in version 62.2.0. We should make sure to test with a version of `setuptools` older than that, but we do already have one such test configured in CI so I think we're set on that front.

Python 3.6 only supports up to `setuptools` 59.6.0, which explains why the tests were failing on 3.6.

src/setuptools_pyproject_migration/__init__.py
diazona: > Learning so many little tricks with this project. :-)

It is definitely surprisingly deep, especially being meant for such a simple task!

diazona:
I think I have this ready now.

One note: I tried changing some of the tests to directly use `Distribution` objects like you did in #46, but they were picking up the wrong version number, so I figured I'd just go ahead with something that works, and hopefully we can fix that issue later.

diazona:
Oh, and just something to be aware of: I expect that this is going to have a merge conflict with #46, which one of us is going to have to fix, but it shouldn't be hard to deal with.

diazona:
Whoops, I forgot to change the assignee when this was ready

sjlongland (approved):

Issue #36, opened 2023-07-17
Support entry points


diazona:
This issue entails adding support for writing the `entry-points`, `gui-scripts`, and `scripts` fields in `pyproject.toml` when they are defined statically in the setuptools configuration.

sjlongland:
So… a relatively complex project where I used two kinds of entry point is this little private project… just created the `debug/dumpmeta` branch which takes the Gist I did earlier, and refines it slightly.

```
{
    "methods": {
        "get_author": "Stuart Longland VK4MSL",
        "get_author_email": "me@vk4msl.id.au",
        "get_classifiers": [],
        "get_cmdline_options": {},
        "get_command_list": [
            "('build', 'build everything needed to install')",
            "('build_py', '\"build\" pure Python modules (copy to build directory)')",
            "('build_ext', 'build C/C++ extensions (compile/link to build directory)')",
            "('build_clib', 'build C/C++ libraries used by Python extensions')",
            "('build_scripts', '\"build\" scripts (copy and fixup #! line)')",
            "('clean', \"clean up temporary files from 'build' command\")",
            "('install', 'install everything from build directory')",
            "('install_lib', 'install all Python modules (extensions and pure Python)')",
            "('install_headers', 'install C/C++ header files')",
            "('install_scripts', 'install scripts (Python or otherwise)')",
            "('install_data', 'install data files')",
            "('sdist', 'create a source distribution (tarball, zip file, etc.)')",
            "('register', 'register the distribution with the Python package index')",
            "('bdist', 'create a built (binary) distribution')",
            "('bdist_dumb', 'create a \"dumb\" built distribution')",
            "('bdist_rpm', 'create an RPM distribution')",
            "('check', 'perform some checks on the package')",
            "('upload', 'upload binary package to PyPI')",
            "('dumpmeta', '(no description available)')",
            "('pyproject', '(no description available)')",
            "('alias', 'define a shortcut to invoke one or more commands')",
            "('bdist_egg', 'create an \"egg\" distribution')",
            "('develop', \"install package in 'development mode'\")",
            "('dist_info', 'DO NOT CALL DIRECTLY, INTERNAL ONLY: create .dist-info directory')",
            "('easy_install', 'Find/get/install Python packages')",
            "('editable_wheel', 'DO NOT CALL DIRECTLY, INTERNAL ONLY: create PEP 660 editable wheel')",
            "('egg_info', \"create a distribution's .egg-info directory\")",
[...]

PR #46, opened 2023-08-05
Issue 36: Implement entrypoint support

sjlongland:
This implements the entry-point support.  (Closes issue #36)


src/setuptools_pyproject_migration/__init__.py
```diff
@@ -91,8 +91,50 @@ def _generate(self) -> Pyproject:
...
+        if "gui_scripts" in entrypoints:
+            pyproject["project"]["gui-scripts"] = entrypoints.pop("gui_scripts")
+
+        # Anything left over gets put in entry-points
+        if entrypoints:
+            pyproject["project"]["entry-points"] = entrypoints
+
         return pyproject
 
+    def _parse_entrypoint(self, entrypoint: str) -> Dict[str, str]:
```
diazona, line 109:
```suggestion
    @staticmethod
    def _parse_entrypoint(entrypoint: str) -> Dict[str, str]:
```

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -91,8 +91,50 @@ def _generate(self) -> Pyproject:
...
+        """
+        Extract the entry-point and name from the string.
+        """
+        # Format: 'hello-world = timmins:hello_world'
```
diazona, line 113: I think it would be useful to have some tests for this method. Personally I would write a file of unit tests like what I did for `_transform_contributors()` in #45, but if nothing else a few doctests would check the main functionality and help illustrate the method's behavior. If you make it a static method you can do this:

```suggestion
        
        >>> WritePyproject._parse_entrypoint("hello-world = timmins:hello_world")
        {'hello-world': 'timmins:hello_world'}
        >>> WritePyproject._parse_entrypoint("hello-world")
        Traceback (most recent call last):
            ...
        ValueError: Entry point 'hello-world' is not of the form 'name = target'
        """
```

(BTW I am partial to using Monty Python references in the examples but that doesn't really matter of course :stuck_out_tongue:)

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -91,8 +91,50 @@ def _generate(self) -> Pyproject:
...
+        if "=" not in entrypoint:
+            raise ValueError("Entry point %r is not of the form 'name = target'" % entrypoint)
+
+        (name, target) = entrypoint.split("=", 1)
+        return {name.strip(): target.strip()}
+
+    def _generate_entrypoints(self) -> Dict[str, Dict[str, str]]:
+        """
+        Dump the entry-points for the project of the given type, if any.
```
diazona, line 122: "given type" seems a bit out of place since this method doesn't accept any parameters. A minimal change might be something like this:

```suggestion
        Dump the entry-points for the project, if any, organized by type.
```

although I personally think it's kind of understood that they will be organized by type so you could just omit that.

```suggestion
        Dump the entry-points for the project, if any.
```

Then again, if you make any of the other changes I'm suggesting below, a different way of describing it might wind up being more appropriate.

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -91,8 +91,50 @@ def _generate(self) -> Pyproject:
...
+        """
+        entrypoints: Dict[str, Dict[str, str]] = {}
```
diazona, line 124: This is somewhat a matter of personal preference, but I would employ the early return pattern here by first checking for the absence of the `entry_points` object:

```suggestion
        if not self.distribution.entry_points:
            return {}

        entrypoints: Dict[str, Dict[str, str]] = {}
```

and then you can skip the `if` statement below. I think it's easier to understand the code this way. (Not that less indentation correlates exactly to easy-to-understand in general, but I think it helps.)

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -91,8 +91,50 @@ def _generate(self) -> Pyproject:
...
+
+        if self.distribution.entry_points:
+            for eptype, raweps in self.distribution.entry_points.items():
+                scripts: Dict[str, str] = {}
+
+                for epstr in raweps:
+                    scripts.update(self._parse_entrypoint(epstr))
+
+                if scripts:
+                    entrypoints[eptype] = scripts
```
diazona, line 134: I think it'd be nice to make the method take `entry_points` as an argument, instead of accessing `self.distribution` directly. That way it would be more clear what the inputs are and easier to reason about the function, e.g. it'll be obvious that it's not accessing random other information from the `Distribution` object.  That would also be reflected in testability, because you could then make it a static method and write some doctests, like with `_parse_entrypoint()`.

Also, if you make `_parse_entrypoint()` return a two-element tuple instead of a dict, you could construct the inner dict a little more efficiently:

```suggestion
            for eptype, raweps in self.distribution.entry_points.items():
                scripts: Dict[str, str] = dict(map(self._parse_entrypoint, raweps))

                if scripts:
                    entrypoints[eptype] = scripts
```

And in fact, I think it would be reasonable to propagate empty lists to empty dicts - that is, if there is a key that exists in `self.distribution.entry_points` with an empty list as its corresponding value, we could just produce an empty list rather than omitting that key. I'm not sure if `setuptools` would ever actually produce an empty list of entry points in that context, but if it does, it presumably means that someone intentionally put a header with no values in `setup.cfg` and if so they would probably want it preserved in `pyproject.toml`. (This would be a good thing to add a test for) Anyway, if you don't worry about empty lists then it simplifies the implementation further:

```suggestion
            for eptype, raweps in self.distribution.entry_points.items():
                entrypoints[eptype] = dict(map(self._parse_entrypoint, raweps))
```

And in fact the entire thing _could_ be condensed to a one-liner, in which case you wouldn't even need to pre-declare `entrypoints`, the whole function body could just be this:

```python
        if not self.distribution.entry_points:
            return {}

        return {
            eptype: dict(map(self._parse_entrypoint, raweps))
            for eptype, raweps in self.distribution.entry_points.items()
        }
```

(or whatever formatting black comes up with) Though this might be going too far down the road of conciseness :stuck_out_tongue: Personally I would stick with the second suggestion.

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -91,8 +91,50 @@ def _generate(self) -> Pyproject:
...
+        # GUI scripts and console scripts go separate in dedicated locations.
+        if "console_scripts" in entrypoints:
+            pyproject["project"]["scripts"] = entrypoints.pop("console_scripts")
+
+        if "gui_scripts" in entrypoints:
+            pyproject["project"]["gui-scripts"] = entrypoints.pop("gui_scripts")
+
+        # Anything left over gets put in entry-points
+        if entrypoints:
+            pyproject["project"]["entry-points"] = entrypoints
```
diazona, line 105: :+1: especially with the comments

diazona (commented):
I wrote up some feedback before realizing you had this marked as a draft, whoops :sweat_smile: but I figured I might as well post it rather than discarding it. Feel free to ignore all my comments for a while if you're still working on this. And as always, all open to discussion.

src/setuptools_pyproject_migration/__init__.py
sjlongland: Yeah, good point… that'll actually make testing the method easier since I don't need to instantiate the class.

src/setuptools_pyproject_migration/__init__.py
sjlongland: Yep, I'm in the process of doing that now.  Figured the unit tests would wait until the morning when I had a clear head. :-)

src/setuptools_pyproject_migration/__init__.py
sjlongland: Yeah, originally it was going to take a parameter, the entry point type (e.g. `console_scripts`, `gui_scripts` or arbitrary module names), and was going to return the modules _of that type_, then I realised I'd have to make another function to return what entry point types are available.

So dropping the parameter and making it iterate over the lot was cleaner, but forgot to clean up the documentation for it.

Making the other method a static method, this too could be simplified a little bit to make unit testing easier.

src/setuptools_pyproject_migration/__init__.py
sjlongland: True, it avoids the march off the right-hand side of the screen.

src/setuptools_pyproject_migration/__init__.py
sjlongland: Yeah, I do use list comprehensions to generate things, I keep forgetting the same trick works with `dict`s.

One thing in the back of my mind is whether there's any possibility of `setuptools` throwing us an empty list of strings as an entry point list.

The suggestions here will yield an empty `dict` as the value rather than omitting it completely.  However, the risk of this is probably quite small -- if I go the second option, we can always add a `if not raweps: continue` to the loop later. :-)

diazona: > One thing in the back of my mind is whether there's any possibility of setuptools throwing us an empty list of strings as an entry point list.

Yeah fair point, but I think if it causes a problem we can fix it like you said. And it's possible the type checker would catch that anyway.

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -91,8 +91,67 @@ def _generate(self) -> Pyproject:
...
+        """
+        Extract the entry point and name from the string.
+
+        .. code-block:: python
+
+            WritePyproject._parse_entrypoint("hello-world = timmins:hello_world")
+            # returns ('hello-world', 'timmins:hello_world')
+
+            WritePyproject._parse_entrypoint("hello-world")
+            # throws ValueError: Entry point 'hello-world' is not of the form 'name = target'
```
diazona, line 120: I would still really recommend making these doctests, unless you have some reason not to. I think we gain quite a bit by making them real runnable tests, e.g. if the behavior of this function ever changes in a way that causes the code to become inaccurate, then if they are doctests it'll be caught by automatic testing, whereas with this syntax we rely on someone trawling through the code and noticing.

```suggestion
        >>> WritePyproject._parse_entrypoint("hello-world = timmins:hello_world")
        {'hello-world': 'timmins:hello_world'}
        >>> WritePyproject._parse_entrypoint("hello-world")
        Traceback (most recent call last):
            ...
        ValueError: Entry point 'hello-world' is not of the form 'name = target'
```

...unless this syntax actually _is_ recognized by doctest? (If so I just haven't seen it before)

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -91,8 +91,67 @@ def _generate(self) -> Pyproject:
...
+
+        :param: entrypoint  The entry point string, of the form
+                            "entrypoint = module:function" (whitespace optional)
+        :returns:           A two-element `tuple`, first element is the entry point name, second element is the target
+                            (module and function name) as a string.
+        :raises ValueError: An equals (`=`) character was not present in the entry point string.
+        """
+        # Format: 'hello-world = timmins:hello_world'
```
diazona, line 128: This comment is no longer needed now that the expected input format is described in the documentation (and the examples).

```suggestion
```

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -91,8 +91,67 @@ def _generate(self) -> Pyproject:
...
+        """
+        if not entrypoints:
+            return {}
+
+        parsedentrypoints: Dict[str, Dict[str, str]] = {}
+
+        for eptype, raweps in entrypoints.items():
+            parsedentrypoints[eptype] = dict(map(cls._parse_entrypoint, raweps))
+
+        return parsedentrypoints
```
diazona, line 153: Definitely optional, but I think this helps with readability (and plus it's consistent with the way we separate `_parse_entrypoint`):

```suggestion
        parsed_entrypoints: Dict[str, Dict[str, str]] = {}

        for eptype, raweps in entrypoints.items():
            parsed_entrypoints[eptype] = dict(map(cls._parse_entrypoint, raweps))

        return parsed_entrypoints
```

Or, I guess what we really _should_ do is make it consistent with the way we spell "entrypoints"/"entry-points"/"entry points" :shrug:

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -91,8 +91,67 @@ def _generate(self) -> Pyproject:
...
+        """
+        # Format: 'hello-world = timmins:hello_world'
+        if "=" not in entrypoint:
+            raise ValueError("Entry point %r is not of the form 'name = module:function'" % entrypoint)
+
+        (name, target) = entrypoint.split("=", 1)
+        return (name.strip(), target.strip())
+
+    @classmethod
+    def _generate_entrypoints(cls, entrypoints: Optional[Dict[str, List[str]]]) -> Dict[str, Dict[str, str]]:
```
diazona, line 136: This is definitely not the most important thing, but I do think a static method makes more sense here. I mean, there's no need for this method to have access to the class except as a way to access `_parse_entrypoint`, but `_parse_entrypoint` is hard-coded in precisely one class whose name is known and can be referenced directly. A class method would make more sense to me if inheritance were in play, i.e. if there were a hierarchy of different classes that could be passed as `cls`, and this method needed to access potentially different implementations based on what `cls` is, but that's not the case here; `WritePyproject` is serving as nothing more than a container for some related functionality, and the two static methods have no dependence on that class whatsoever.

In fact it could be argued that it would make more sense to put the two methods as module-level functions. I thought about suggesting that in my initial review, but I didn't do so because the methods are never used outside the class and I thought it made sense to minimize the scope in which they're visible, just to make the code easier to follow. (But this is definitely debatable; perhaps they _do_ make more sense as module-level functions.)

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -91,8 +91,67 @@ def _generate(self) -> Pyproject:
...
+        """
+        Dump the entry-points given, if any.
```
diazona, line 138: I think it'd help a bit to be consistent about whether it's spelled "entrypoints", "entry-points", or "entry points", although I'm not sure which one is best. (TBH I'm not even sure if setuptools itself is consistent about this... if not, I guess this comment becomes kind of moot)

tests/test_entrypoint.py
```diff
@@ -0,0 +1,239 @@
...
+        },
+        "gui_scripts": {
+            "dead-parrot": "montypython.sketch:petshop",
+            "shrubbery": "montypython.holygrail:knightswhosayni",
+        },
+        "project.plugins": {
+            "babysnatchers": "montypython.somethingcompletelydifferent:babysnatchers",
+            "eels": "montypython.somethingcompletelydifferent:eels",
+        },
+    }
```
diazona, line 82: For longer statements like this, I find it valuable to put the expected result in a variable, and maybe even put the inputs in a variable as well, because when a test fails in pytest, it will print out the failing `assert` statement on the console, and I think it's useful to have that printed statement be short so that it's easy to identify what failed. (Although pytest does abbreviate some things so maybe the output would be reasonable regardless... I guess you could try making some change that would cause this test to fail and check what happens.)

P.S. Nice with the Monty Python references :smile:

tests/test_entrypoint.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+# Test correct placement of the parsed endpoints
```
diazona, line 85: I noticed that the rest of the tests in this file are, I guess, integration tests (i.e. testing the overall operation of the plugin rather than testing individual functions like `_generate_entrypoints()`), and it got me thinking that personally I would probably split those into a separate file to make the distinction between the two kinds of tests more clear, and also to make it easier to run only one or the other kind. (Integration tests take longer, if nothing else) But honestly that's just my personal preference and I don't know if it's really the best way to do it... Anyway, I am fine to go forward either way, but it's something to think about before finalizing this PR.

It is definitely a great thing to _have_ tests that just feed a manually-constructed `Distribution` object to `WritePyproject` though. I should have written more of those in my previous work, and I definitely want to follow your lead and lean more heavily on them in the future.

tests/test_entrypoint.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+    cmd = WritePyproject(
+        Distribution(
+            dict(
+                name="TestProject",
+                version="1.2.3",
+            )
+        )
+    )
+    project = cmd._generate()
```
diazona, line 100: Minor thing, but in other tests I've been using `result` as the name of the data structure that gets returned from `_generate()`, and it might be handy to stay consistent on that. (I mean, `result` isn't a great name, and we could definitely change it, but the consistency is what matters.) It might throw someone off a bit to use `project` when that is also the name of a fixture that we use prominently in other tests.

tests/test_entrypoint.py
```diff
@@ -0,0 +1,239 @@
...
+
+def test_parse_entrypoint_missing_eq():
+    """
+    Test that a string lacking a '=' is refused.
+    """
+    try:
+        WritePyproject._parse_entrypoint("ep project.module:target")
+        assert False, "Should not have been accepted as valid"
+    except ValueError as e:
+        assert str(e).endswith(" is not of the form 'name = module:function'")
```
diazona, line 37: pytest has a [standard way to ensure that an exception gets raised](https://docs.pytest.org/en/7.3.x/reference/reference.html#pytest-raises):

```suggestion
    with pytest.raises(ValueError, match=" is not of the form 'name = module:function'$"):
        WritePyproject._parse_entrypoint("ep project.module:target")
```

(don't forget to `import pytest`)

diazona (changes requested):

tests/test_entrypoint.py
sjlongland: Indeed, I guess I was thinking about the object data type (`_generate()` returns a `Project`; and `project` here is an instance of it), but as we're using that for a fixture, perhaps the more generic `result` is better.

src/setuptools_pyproject_migration/__init__.py
sjlongland: Yeah, not completely sure myself.  The property on the `Distribution` object is called `entry_points`; whether you read the `_` as being a space or an underscore is up for debate.

In the docs, it's not hyphenated in the title, it's a space: https://setuptools.pypa.io/en/latest/userguide/entry_point.html -- so yeah, maybe the hyphen can go.

src/setuptools_pyproject_migration/__init__.py
sjlongland: Yeah, I was avoiding hard-coding the class, but you are right, that's the only reason why it's a class method.

If we made them module-level functions, we can still unit test them individually: `__all__` doesn't stop us importing them and calling them directly (it only specifies what `*` gives you if you do a `from ${module} import *`).

Probably less clutter for the class though, and less probability that we'll confuse people later.

src/setuptools_pyproject_migration/__init__.py
sjlongland: True, I vaguely recall there being conventions around using underscores in variable names… but seems it's on _packages_ where underscores are discouraged, not variables.

https://peps.python.org/pep-0008/#package-and-module-names

So yeah, when I next get a moment… a underscore should be added here.

tests/test_entrypoint.py
sjlongland: Yep, was definitely wracking the brain there regarding some references.  It was after all, how the programming language got its name. :-)

I can refactor this, no big deal.

tests/test_entrypoint.py
sjlongland: Yeah, I sort of grouped these all in since I'm specifically looking at the entry point outputs (and ignoring everything else).

I think the big distinction is that most of the tests of this nature are invoking the TOML serialiser, whereas this is comparing the parts of the `dict` directly.  Those are definitely "integration" tests since about the only thing we _don't_ do is generate a dummy `setup.py` and run it in a sub-process, we're not far removed from that though.

As for time taken:

```
======================================================= test session starts =======================================================
platform linux -- Python 3.11.4, pytest-7.4.0, pluggy-1.2.0 -- /home/stuartl/projects/setuptools-pyproject-migration/env/bin/python3
cachedir: .pytest_cache
rootdir: /home/stuartl/projects/setuptools-pyproject-migration/work
configfile: pytest.ini
plugins: cov-4.1.0, console-scripts-1.4.1
collected 10 items                                                                                                                

tests/test_entrypoint.py::test_parse_entrypoint_happypath PASSED                                                            [ 10%]
tests/test_entrypoint.py::test_parse_entrypoint_happypath_whitespace PASSED                                                 [ 20%]
tests/test_entrypoint.py::test_parse_entrypoint_missing_eq PASSED                                                           [ 30%]
tests/test_entrypoint.py::test_generate_entrypoints_none PASSED                                                             [ 40%]
tests/test_entrypoint.py::test_generate_entrypoints PASSED                                                                  [ 50%]
tests/test_entrypoint.py::test_generate_noentrypoints PASSED                                                                [ 60%]
tests/test_entrypoint.py::test_generate_clionly PASSED                                                                      [ 70%]
tests/test_entrypoint.py::test_generate_guionly PASSED                                                                      [ 80%]
tests/test_entrypoint.py::test_generate_misconly PASSED                                                                     [ 90%]
tests/test_entrypoint.py::test_generate_all_entrypoints PASSED                                                              [100%]

======================================================= 10 passed in 0.12s ========================================================
```

This machine is definitely no spring chicken:
```
RC=0 stuartl@rikishi ~/projects/setuptools-pyproject-migration/work $ grep '^model name' /proc/cpuinfo | head -n 1
model name      : Intel(R) Core(TM) i5-3320M CPU @ 2.60GHz
```

I do note the `test_everything` test does take a lot longer (almost 25 seconds), not quite sure why at this point, but I'm not overly bothered either… 30 seconds is not long to wait for test results.

src/setuptools_pyproject_migration/__init__.py
diazona: In that case I'd go with "entry points" for docs and `entry_points` for code.

tests/test_entrypoint.py
diazona: Sure, fair enough.

Personally I find 25 seconds long enough to break concentration, so I do tend to avoid that test when I'm quickly iterating on things. But yeah, it's not the biggest deal. Perhaps it takes a while because it's installing dependencies.

diazona:
Testing something:

Closes #36

diazona:
I think it might be necessary to include "Closes #36" in the issue description to get the issue to be automatically closed when this is merged. (Putting it in a comment doesn't seem to be enough)

src/setuptools_pyproject_migration/__init__.py
sjlongland: Yeah, I guess I was trying to think how the Sphinx docs engine would render it or how to mark it up to indicate to the documentation system "this is example code", and `>>>` isn't valid Python syntax.

No biggie though, we'll give the "raw" formatting method a try and see how it goes.

diazona: Hmm, good point, I totally forgot to consider how Sphinx renders this. Whoops :sweat_smile: But there must be a way to keep the `>>>` prefix (which AFAIK doctest needs to recognize it as an executable test) while also ensuring that Sphinx will produce something reasonable. I mean, I'm pretty sure I've seen other projects do both. If it doesn't just work as-is, off the top of my head I'm guessing there might be a language like "ipython" or "python prompt" or something that you can use for the code block, which will allow `>>>`. (That's why doctest uses `>>>`, because it's supposed to look like an interactive Python session.)

Anyway, I'm fine either way: we can go ahead with "raw" formatting and see what Sphinx does, and if it doesn't come out looking good we'll fix it up in another PR; or if you prefer, feel free to figure out how to make them both work together, and we'll do that. We can fix up any doctests that have already been merged in another PR.

diazona:
Don't forget to resolve the merge conflicts. (I don't care whether you do it by rebasing or by merging `main` into your branch)

Feel free to resolve comment threads that have been addressed or are otherwise no longer needed if you like, but I will also come back to this sometime in the next day or so and review all the comment threads, resolving as needed. Right now I only had time for a quick peek. (I'm supposed to be at work lol, but it's quiet at the moment)

src/setuptools_pyproject_migration/__init__.py
diazona: ...actually Sphinx doesn't even render documentation for private methods (starting with `_`), it seems. I'm sure there's a configuration setting for that but to me it makes sense to keep it as-is.

I still kind of like _having_ the documentation even if it's not actually rendered. It's handy for people reading the code.

src/setuptools_pyproject_migration/__init__.py
sjlongland: Yep, I've just revised everything so we're using `entry_points` in code, "entry points" in documentation.  The only place where we use `entry-points` is in the TOML output.

src/setuptools_pyproject_migration/__init__.py
diazona: Funnily enough, I actually encountered a couple of cases at work this week where I got confused about the purpose of a function that was at top-level scope even though it was only used in a specific local scope. So that's got me leaning further toward a static method being the right solution for these - it keeps them closer to where they're actually used. But I don't think it's worth the trouble to change back at this point; I'm just making this point as something to keep in the back of our mind in the future. If nothing else, it is annoyingly inconsistent to have some helper functions as static methods and others not, so some kind of refactoring will be necessary at some point.

Maybe it will even make sense to break out all the functionality related to entry points into a separate class or even a separate module or something :shrug: but again, not something to deal with now.

diazona (approved):

PR #48, opened 2023-08-08
Split linting/static checks apart from pytest

### LINKED ISSUE #39: Separate static checks from pytest
I want to separate the static code checks (black, mypy, ruff) so that they don't run as part of pytest. Instead they should run either as part of pre-commit, or as their own standalone tox environments. We only need to run these checks once, not for every Python version as we do with pytest, and besides the dependencies needed to run these tests as part of pytest are breaking compatibility with Python 3.6.

diazona:
Static checks are already run by pre-commit, which gets called from CI, so there is very little I had to do to split them out of pytest. I just removed the pytest plugins that enable those checks.

I added a tox environment to allow people to run the static checks through tox in case they don't want to install and run pre-commit directly for some reason.

Closes #39

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,17 +42,9 @@ testing =
 	pytest >= 6
 	pytest-console-scripts >= 1.2
 	pytest-checkdocs >= 2.4
-	pytest-black >= 0.3.7; \
-		# workaround for jaraco/skeleton#22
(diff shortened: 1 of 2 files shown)


diazona:
I enabled auto-merge on this commit for fun, let's see how it works :stuck_out_tongue:

PR #54, opened 2023-08-12
Add contribution guidance in CONTRIBUTING.md

diazona:
I'm making an attempt at adding some contribution guidelines for anyone interested in working on the project. This is not the most comprehensive but we can always edit it later as we think of better ways to express the guidelines and/or get feedback from contributors about what would be useful.

Suggestions welcome on this....


sjlongland (approved):
Yep, it seems to cover most bases.  This no doubt will change as needs are identified.

PR #55, opened 2023-08-12
Fix or ignore all warnings

diazona:
I fixed the source of one warning related to the code in this project, and ignored some others that are issued by setuptools/distutils code. I think ignoring is the right call there because the whole purpose of this project is to process legacy configuration so we are naturally going to use some deprecated features, but I wouldn't mind having that double-checked.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -27,3 +27,9 @@ filterwarnings=
 	## end upstream
 
 	ignore:The distutils package is deprecated and slated for removal:DeprecationWarning
+
+	ignore:setuptools.installer (is|and fetch_build_eggs are) deprecated::setuptools
+	ignore:'encoding' argument not specified::setuptools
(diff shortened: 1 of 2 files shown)


sjlongland:
Yep, the idea of this project is people will be able to run the old `setup.py` one last time… and finally retire it from service.

So if `setuptools`/`distutils` squawks… no big deal.  I think some of the warnings were file encoding issues… if memory serves, Python historically used ISO-8859-1 (aka `latin1`) encoding, and moved to UTF-8 with Python 3.

diazona:
This one looks ready. I got rid of all the warnings I saw showing up in CI, not because it was really necessary to do so, but I figured I might as well get them all as long as I was looking into warnings in the first place.

We could configure pytest to upgrade all (unignored) warnings to errors if we want to, although it certainly doesn't have to happen as part of this PR.

sjlongland (approved):

PR #57, opened 2023-08-14
Support the description field

diazona:
This PR adds support and a few tests for the `description` field. It's pretty straightforward.

### DIFF
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -201,6 +201,11 @@ class WritePyproject(setuptools.Command):
         if classifiers:
             pyproject["project"]["classifiers"] = classifiers
 
+        description: str = dist.get_description()
+        # "UNKNOWN" is used by setuptools<62.2 when the description in setup.cfg is empty or absent
+        if description and description != "UNKNOWN":
+            pyproject["project"]["description"] = description
+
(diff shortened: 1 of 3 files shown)


diazona:
The dreaded "UNKNOWN" strikes again :laughing: I guess that's what I get for saying it was straightforward

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -201,6 +201,11 @@ def _generate(self) -> Pyproject:
         if classifiers:
             pyproject["project"]["classifiers"] = classifiers
 
+        description: str = dist.get_description()
+        # "UNKNOWN" is used by setuptools<62.2 when the description in setup.cfg is empty or absent
+        if description and description != "UNKNOWN":
```
sjlongland, line 206: I'm beginning to wonder how many other places where they use this `"UNKNOWN"` sentinel value.  Looks like not just author information.

sjlongland (approved):

src/setuptools_pyproject_migration/__init__.py
diazona: Yeah apparently not. I wouldn't be surprised if it shows up in like half the fields... But I still think it's likely easier to just deal with them as we encounter them. Or at least I can't think of a cleaner way.

PR #58, opened 2023-08-14
Support readme files

diazona:
This pull request adds support for readme files, populated by what setuptools calls the long description. This one is a bit more complicated because the naive way to get the field value, `get_long_description()`, reads the content of the readme file if one is specified, but we need the actual filename to write into `pyproject.toml`, so I went a little deep into the `Distribution` metadata to find the raw value.

The other notable thing about this bit of code is that it will actually write the long description to a file if it was hard-coded in the setuptools parameters. This is the first case in which the plugin generates something other than the content of a `pyproject.toml` file.

### DIFF
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -1,7 +1,9 @@
 import itertools
+import mimetypes
 import setuptools
 import sys
 import tomlkit
+import warnings
 from pep508_parser import parser as pep508
 from typing import Dict, List, Optional, Tuple, Type, Union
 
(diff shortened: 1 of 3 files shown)


diazona:
This caused a failure in local testing lol

```console
$ python3.6
Python 3.6.15 (default, Jan  3 2022, 03:00:17) 
[GCC 10.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import mimetypes
>>> mimetypes.guess_extension("text/plain")
'.bat'
```

:exploding_head: :facepalm: :facepalm:

Apparently [it affects Python <3.8](https://bugs.python.org/issue1043134)

sjlongland:
> This caused a failure in local testing lol
> 
> ```
> $ python3.6
> Python 3.6.15 (default, Jan  3 2022, 03:00:17) 
> [GCC 10.3.0] on linux
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import mimetypes
> >>> mimetypes.guess_extension("text/plain")
> '.bat'
> ```
> 
> 🤯 🤦 🤦
> 
> Apparently [it affects Python <3.8](https://bugs.python.org/issue1043134)

`.bat` eh?  Blast from the past.  Given it's implicated in old Python releases, that are no longer maintained, I don't see them fixing it upstream anytime soon.

diazona:
I know I had mentioned adding tests that use `setup.py` directly in a separate discussion, but I kind of want to just get this merged without them and go back and add such tests later. (Even though the `readme` handling really _should_ have such tests, because it digs into the setuptools data structures)

sjlongland (approved):

sjlongland:
That makes sense… I would have thought `setuptools` would behave the same way in both cases, but it seems you can get different behaviour depending on which method a project uses.

We didn't know this at the start… we've only just discovered that little gem.  We'll come back to the inline case later I think.

PR #60, opened 2023-08-19
Bugfix/setuptools circular import workaround

sjlongland:
Cherry-pick of the fixes from PR#59 to address issues in `setuptools`.


diazona (approved):
Works for me. I'd just suggest creating an issue as a reminder to remove the restriction when we can. (As for whether that means removing it permanently because there's an upstream fix, or just switching it out for some less onerous restriction, we can figure that out as we go)

sjlongland:
For what it's worth, `setuptools` version 68 is still used in Python 3.12… it's earlier versions the limit applies to -- largely because of the `docs` target which breaks.

PR #66, opened 2023-08-21
Support URLs

### LINKED ISSUE #35: Support urls
This issue entails adding support for writing the `urls` field in `pyproject.toml` when it is defined statically in the setuptools configuration.

diazona:
This PR adds support for URLs.

In its current form, it only copies over setuptools' `project_urls` field into the `urls` field of the output dict, but I suspect it's better to do something with the `url` and `download_url` fields if they exist. So I want to hold off on merging this for a while, to give us some time to think about it.

Closes #35

### DIFF
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -257,6 +257,10 @@ class WritePyproject(setuptools.Command):
         if classifiers:
             pyproject["project"]["classifiers"] = classifiers
 
+        urls: Dict[str, str] = dist.metadata.project_urls
+        if urls:
+            pyproject["project"]["urls"] = urls
+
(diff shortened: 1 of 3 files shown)


diazona:
I'm still not sure what to do when there are URLs in the `url` and/or `download_url` fields of `Distribution`, but I think the best course of action at the moment might be to just merge this and then do some real-world research to figure out how to handle those. It may well depend on what usage patterns really exist in the wild, e.g. whether people typically use `project_urls` as a complement to `url` and `download_url` or as a replacement for them.

sjlongland:
Well, if there's a little manual clean-up afterwards, I don't see this being a big problem.  The primary goals were things like dependencies, where you really do want to keep things in sync.

sjlongland (approved):

PR #67, opened 2023-08-21
Add a badge reflecting the status of the documentation build

diazona:
I noticed that we don't have a simple indicator of whether the docs are building successfully on readthedocs or not. readthedocs imposes more stringent requirements than the "native" documentation build that we run in tox, so it's very possible that it could be failing even though our own documentation build is passing, and we need a way to be notified of that. This commit adds the badge from readthedocs which will fulfill that purpose.


sjlongland (approved):

diazona:
...the docs build started passing again right as soon as we merged this PR :laughing:

PR #75, opened 2023-09-04
Add WIP repo status badge

diazona:
This badge text comes from https://www.repostatus.org, and appears to be a fairly well-defined way to indicate the status of a project. It's similar to the "Development Status" trove classifiers, but is more visible and not specific to Python.

This is not super important but seems like a nice bit of added value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,15 @@
 # setuptools-pyproject-migration
 
+<!-- markdownlint-disable MD013 -->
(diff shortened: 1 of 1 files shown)


diazona:
Looks like [a runner error](https://github.com/actions/runner/issues/1923) is blocking tests from passing

sjlongland:
I've given the CI system a sharp prod, seems to be going again.

Issue #32, opened 2023-07-17
Support license


diazona:
This issue entails adding support for writing the `license` field in `pyproject.toml` when it is defined statically in the setuptools configuration.

sjlongland:
So there are two mechanisms for including a license, inline licenses and external files.  `pyproject.toml` only supports the latter, so much like the `long_description` (issue #33), we'll need to write a file out and reference it in the `pyproject.toml`.

This suggests in the medium-term, we should also emit `pyproject.toml` to a file rather than dumping it to standard output, and/or provide a command switch to specify *which* file we output to standard output… but this will require careful thought.

sjlongland:
The inline case looks like this:

`setup.py` source:

```python
import setuptools

setuptools.setup(
    name="mypackage",
    version="1.2.3",
    license="""Mock License 1.0

This is a dummy license text to check how setuptools represents
it in the metadata so it can be exposed to pyproject.toml."""
)
```

`dumpmeta` output:
```
RC=0 stuartl@rikishi ~/projects/setuptools-pyproject-migration/experiments/license/inline $ python setup.py dumpmeta
running dumpmeta
{
    "methods": {
…snip…
        "get_license": "Mock License 1.0\n\nThis is a dummy license text to check how setuptools represents\nit in the metadata so it can be exposed to pyproject.toml.",
…snip…
    },
    "properties": {
…snip…
        "metadata": "<distutils.dist.DistributionMetadata object at 0x7f1c18a81410>",
…snip…
    }
}
```

PR#58 actually inspects that `metadata` structure, so might tweak `dumpmeta` to inspect that more closely.

diazona:
Oh I actually made some local improvements on the `dumpmeta` branch while working on readme support, along very similar lines to what you mentioned - I can push those when I get home later today.

diazona:
> This suggests in the medium-term, we should also emit `pyproject.toml` to a file rather than dumping it to standard output, and/or provide a command switch to specify which file we output to standard output… but this will require careful thought.

Yeah agreed. I had a feeling for a while that we might wind up going in that direction, but I thought it's something we didn't necessarily need to figure out until later. Printing to stdout was just easier 😛

Of course, one of the complications with writing to `pyproject.toml` is that it might already exist and we'd have to merge the existing contents with whatever we get from setuptools. Hopefully that won't be too hard, but it is some extra complexity. I think it would be fair to leave that out of an initial alpha/beta release.

[...]

PR #78, opened 2023-09-09
Add license field support

sjlongland:
One gripe I have is that `tomlkit` makes a new section, rather than doing this as an inline table.

So instead of this:

```toml
license = { text = "My License" }
```

we get:

```toml
[project.license]
text = "My License"
```

The latter may work better visually for multi-line licenses (the field can store a full custom license agreement if none of the "standard" ones suit), and _should_ be equivalent in terms of TOML behaviour.

But it doesn't match the examples shown in Python packaging docs.

It's not obvious how I coax `tomlkit` to emit the other format.

Reading the TOML spec, the two formats are seemingly identical in terms of the data represented, so I'll leave it as-is for now.  Closes Issue #32.

### DIFF
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -245,6 +245,11 @@ class WritePyproject(setuptools.Command):
             "version": dist.get_version(),  # TODO try to reverse-engineer dynamic version
         }
 
+        # In older setuptools releases, unspecified license text is replaced with "UNKNOWN"
+        license_text = dist.get_license()
+        if license_text and (license_text != "UNKNOWN"):
+            pyproject["project"]["license"] = {"text": license_text}
+
         authors: List[Contributor] = self._transform_contributors(dist.get_author(), dist.get_author_email())
         if authors:
             pyproject["project"]["authors"] = authors
--- a/tests/test_license.py
+++ b/tests/test_license.py
@@ -1,0 +1,241 @@
+"""
+Tests of the ``license`` field in ``pyproject.toml``, which can be populated
+from setuptools' ``license`` field.
+"""
+
+import textwrap
+
+# setuptools *ONLY* supports bare license strings in the `license` field.  There exists a
+# `license_files` and a deprecated `license_file` field.  The latter simply raises a warning and the
+# content is prepended to `license_files`.  Neither populates the `license` property.
+# `license` and `license_files` are the subject of a PEP: https://peps.python.org/pep-0639/  If accepted, likely
+# `license` will be deprecated in favour of a `license-expression` (for SPDX descriptors).
+
+
+def test_license_setupcfg_expression_string(project) -> None:
+    license_string = "MIT"
+    setup_cfg = f"""\
+[metadata]
+name = test-project
+version = 0.0.1
+license = {license_string}
+"""
+
+    project.setup_cfg(setup_cfg)
+    project.setup_py()
+    result = project.generate()
+    license_data = result["project"]["license"]
+    assert license_data == {"text": license_string}
+
+
+def test_license_setupcfg_fulltext_string(project) -> None:
+    license_string = """\
+Python License, Version 2
+
+1. This LICENSE AGREEMENT is between the Python Software Foundation
+(“PSF”), and the Individual or Organization (“Licensee”) accessing and
+otherwise using this software (“Python”) in source or binary form and
+its associated documentation.
+
+2. Subject to the terms and conditions of this License Agreement, PSF
+hereby grants Licensee a nonexclusive, royalty-free, world-wide
+license to reproduce, analyze, test, perform and/or display publicly,
+prepare derivative works, distribute, and otherwise use Python
+alone or in any derivative version, provided, however, that PSF’s
+License Agreement and PSF’s notice of copyright, i.e., “Copyright (c)
+2001, 2002, 2003, 2004, 2005, 2006 Python Software Foundation; All Rights
+Reserved” are retained in Python alone or in any derivative version
+prepared by Licensee.
+
+3. In the event Licensee prepares a derivative work that is based on
+or incorporates Python or any part thereof, and wants to make
+the derivative work available to others as provided herein, then
+Licensee hereby agrees to include in any such work a brief summary of
+the changes made to Python.
+
+4. PSF is making Python available to Licensee on an “AS IS”
+basis. PSF MAKES NO REPRESENTATIONS OR WARRANTIES, EXPRESS OR
+IMPLIED. BY WAY OF EXAMPLE, BUT NOT LIMITATION, PSF MAKES NO AND
+DISCLAIMS ANY REPRESENTATION OR WARRANTY OF MERCHANTABILITY OR FITNESS
+FOR ANY PARTICULAR PURPOSE OR THAT THE USE OF PYTHON WILL NOT
+INFRINGE ANY THIRD PARTY RIGHTS.
+
+5. PSF SHALL NOT BE LIABLE TO LICENSEE OR ANY OTHER USERS OF PYTHON
+FOR ANY INCIDENTAL, SPECIAL, OR CONSEQUENTIAL DAMAGES OR LOSS AS
+A RESULT OF MODIFYING, DISTRIBUTING, OR OTHERWISE USING PYTHON,
+OR ANY DERIVATIVE THEREOF, EVEN IF ADVISED OF THE POSSIBILITY THEREOF.
+
+6. This License Agreement will automatically terminate upon a material
+breach of its terms and conditions.
+
+7. Nothing in this License Agreement shall be deemed to create any
+relationship of agency, partnership, or joint venture between PSF and
+Licensee. This License Agreement does not grant permission to use PSF
+trademarks or trade name in a trademark sense to endorse or promote
+products or services of Licensee, or any third party.
+
+8. By copying, installing or otherwise using Python, Licensee
+agrees to be bound by the terms and conditions of this License
+Agreement.
+
+BEOPEN.COM LICENSE AGREEMENT FOR PYTHON 2.0
+——————————————-
+
+BEOPEN PYTHON OPEN SOURCE LICENSE AGREEMENT VERSION 1
+
+1. This LICENSE AGREEMENT is between BeOpen.com (“BeOpen”), having an
+office at 160 Saratoga Avenue, Santa Clara, CA 95051, and the
+Individual or Organization (“Licensee”) accessing and otherwise using
+this software in source or binary form and its associated
+documentation (“the Software”).
+
+2. Subject to the terms and conditions of this BeOpen Python License
+Agreement, BeOpen hereby grants Licensee a non-exclusive,
+royalty-free, world-wide license to reproduce, analyze, test, perform
+and/or display publicly, prepare derivative works, distribute, and
+otherwise use the Software alone or in any derivative version,
+provided, however, that the BeOpen Python License is retained in the
+Software, alone or in any derivative version prepared by Licensee.
+
+3. BeOpen is making the Software available to Licensee on an “AS IS”
+basis. BEOPEN MAKES NO REPRESENTATIONS OR WARRANTIES, EXPRESS OR
+IMPLIED. BY WAY OF EXAMPLE, BUT NOT LIMITATION, BEOPEN MAKES NO AND
+DISCLAIMS ANY REPRESENTATION OR WARRANTY OF MERCHANTABILITY OR FITNESS
+FOR ANY PARTICULAR PURPOSE OR THAT THE USE OF THE SOFTWARE WILL NOT
+INFRINGE ANY THIRD PARTY RIGHTS.
+
+4. BEOPEN SHALL NOT BE LIABLE TO LICENSEE OR ANY OTHER USERS OF THE
+SOFTWARE FOR ANY INCIDENTAL, SPECIAL, OR CONSEQUENTIAL DAMAGES OR LOSS
+AS A RESULT OF USING, MODIFYING OR DISTRIBUTING THE SOFTWARE, OR ANY
+DERIVATIVE THEREOF, EVEN IF ADVISED OF THE POSSIBILITY THEREOF.
+
+5. This License Agreement will automatically terminate upon a material
+breach of its terms and conditions.
+
+6. This License Agreement shall be governed by and interpreted in all
+respects by the law of the State of California, excluding conflict of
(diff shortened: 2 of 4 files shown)


tests/test_license.py
```diff
@@ -0,0 +1,25 @@
+"""
+Tests of the ``license`` field in ``pyproject.toml``, which can be populated
+from setuptools' ``license`` field.
+"""
+
+# setuptools *ONLY* supports bare license strings in the `license` field.  There exists a
+# `license_files` and a deprecated `license_file` field.  The latter simply raises a warning and the
+# content is prepended to `license_files`.  Neither populates the `license` property.
+# `license_files` is the subject of a PEP: https://peps.python.org/pep-0639/
```
diazona, line 9: I think it would be useful to add a test that sets the `license_files` field in this file, and probably one for `license_file` as well, because IMO it's a good idea to document and constraint the expected behavior of the plugin when one of those fields is present, even if all it shows is that they don't (yet) show up in `pyproject.toml`.

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -239,6 +239,11 @@ def _generate(self) -> Pyproject:
             "version": dist.get_version(),  # TODO try to reverse-engineer dynamic version
         }
 
+        # In older setuptools releases, unspecified license text is replaced with "UNKNOWN"
+        license_text = dist.get_license()
+        if license_text and (license_text != "UNKNOWN"):
+            pyproject["project"]["license"] = dict(text=license_text)
```
diazona, line 245: Almost missed this one, but the same deal about dict literal syntax here. Like the other case, this is a very optional change.

```suggestion
            pyproject["project"]["license"] = {"text": license_text}
```

diazona (changes requested):
Sorry it took me a little while to get to this! I offered a bit of feedback, though none of the changes are strictly essential.

src/setuptools_pyproject_migration/__init__.py
sjlongland: Yeah, I was being lazy I guess. :-)

tests/test_license.py
sjlongland: I didn't see a lot of point in a test that shows `license_files` (and by extension, `license_file`) are consigned to `/dev/null`.  I'm thinking this probably belongs in a separate `test_license_files.py`.

tests/test_license.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_license_setupcfg_string(project) -> None:
+    license_string = "Mock License v1.0"
```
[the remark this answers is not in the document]
sjlongland: I think it can be either, if a "standard" license fits, you use the license name, otherwise if it's something special you spell out what that special license is.

https://packaging.python.org/en/latest/specifications/core-metadata/#license

i.e.
```python
license_string = """This software may only be obtained by sending the
author a postcard, and then the user promises not
to redistribute it."""
```

or

```python
license_text="GPL version 3, excluding DRM provisions"
```

`MIT` was chosen since it matches the classifiers used elsewhere in the tests.

tests/test_setup_command.py
```diff
@@ -91,6 +91,7 @@ def test_everything(project, runner) -> None:
 maintainer = David Zaslavsky, Stuart Longland
 maintainer_email = diazona@ellipsix.net, me@vk4msl.com
 description = A dummy project with a sophisticated setup.cfg file
+license = MIT
```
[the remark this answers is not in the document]
sjlongland: Yeah, I used `MIT` in the example config here since that matches what's in the classifiers below in the same dummy project `setup.cfg`.

The dual-purpose of this field is confusing to say the least.

tests/test_license.py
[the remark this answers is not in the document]
diazona: Ah I see, thanks for looking that up.

All I want to head off is that, if someone happens to be looking at our code and tests as an example of what to put in the `license` field of their `setup.cfg` (it could happen :shrug:), and in every case they see something that looks like the name of a license, they might come away with the impression that the field is _supposed_ to contain the name of the license and not something else. Yes, it's technically better to read the official documentation but I couldn't realistically fault anyone for looking at an example and copying its style - certainly I've done that often enough myself.

diazona:
Oh I meant to reply about the formatting in the TOML output as well. It would definitely be nice if we could make it look like the Python packaging docs, but given that both forms are equivalent, I definitely don't see it as something particularly important. Maybe in the future if we find a way we can modify the code accordingly. (And if someone cares enough, they can make a manual edit to their `pyproject.toml`.)

tests/test_license.py
```diff
@@ -0,0 +1,25 @@
...
+name = test-project
+version = 0.0.1
+license = {license_string}
+"""
+
+    project.setup_cfg(setup_cfg)
+    project.setup_py()
+    result = project.generate()
+    license_data = result["project"]["license"]
+    assert license_data == dict(text=license_string)
```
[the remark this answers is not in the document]
sjlongland: Yep, just amended that now.

tests/test_license.py
sjlongland: I've added a couple for `license_files`; one for literal file names, the other for file path globs.

I've omitted `license_file`, since it's deprecated in `setuptools` now, _and_ it copies the data into `license_files` for us, so if we support the latter, implicitly the former will work too.

Both these tests are marked `xfail` as we do not yet support this feature.

tests/test_license.py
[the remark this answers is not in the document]
sjlongland: Yep… well I've put tests that check both ways actually work as expected.  Since both ways are supported upstream.

PEP-639 should clean this up though.  The core metadata will have a `License-Expression` field (stored as `license` in the TOML) that'll be a SPDX license expression.  So in the future, we will need to validate whether this is a valid SPDX license expression, and if it is, it goes into `license`, otherwise, we write it out to a file and reference it in `license_files`.

So the current `license` "table" will change if PEP-639 is ratified in the form it takes today.

tests/test_license_files.py
```diff
@@ -0,0 +1,52 @@
+"""
+Skeleton tests for `license_files` from `setup.cfg`.  This is a `setuptools`
+extension which is not yet ratified by the Python community.
+
+`license_files` is the subject of a PEP: https://peps.python.org/pep-0639/
```
diazona, line 5: I think this should be restructured text markup, although we don't actually generate documentation for these files (yet?) so it's not a blocker

tests/test_license_files.py
```diff
@@ -0,0 +1,52 @@
...
+
+import pytest
+
+
+@pytest.mark.xfail
+def test_license_files_setupcfg_globs(project) -> None:
+    """
+    Test that file names that contain glob characters are added as globs.
+    """
+    license_files = ["LICEN[CS]E*", "COPYING*", "NOTICE*", "AUTHORS*"]
```
diazona, line 16: I think this is the default value of the field, right? It might be better to use a non-default value so the test can tell if it's just using the default when it shouldn't.

tests/test_setup_command.py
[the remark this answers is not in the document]
diazona: Yeah... The thing is, the classifiers have to come from a fixed list, so we have to use a real license there (if there's going to be a license classifier at all), but this has no such restriction.

Still, it could be something like "MIT license could go here" as far as I'm concerned.

tests/test_license_files.py
```diff
@@ -0,0 +1,52 @@
...
+    project.setup_py()
+    result = project.generate()
+
+    assert result["project"]["license_files"] == {"globs": license_files}
+
+
+@pytest.mark.xfail
+def test_license_files_setupcfg_files(project) -> None:
+    """
+    Test that file names that list explicitly existing files are added as files.
```
diazona, line 36: Are the files actually supposed to exist in this test? I wasn't sure if you meant to add code that creates them.

diazona (commented):
I've got some stuff going on this weekend so I haven't gotten to leave a full review, but hopefully these comments help move things along 🙂

tests/test_license_files.py
sjlongland: Yeah, good point… next time I get a moment I'll fix that.

tests/test_license_files.py
sjlongland: Yeah, we will have to figure out some way to "expose" the content of those files.

One possibility is that we have a stub-able `write_file` method that we simply replace in our unit tests to dump the file's name and content somewhere for comparison purposes.

I think the `long_description` had a similar problem did it not?

diazona: Wait, I may have misunderstood: is the intent here that the files should already exist when the plugin runs, or that the plugin should create the files? I read the docstring to imply the former.

tests/test_license.py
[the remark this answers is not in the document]
diazona: ...I suppose that works. I mean, I wasn't claiming that it actually matters for the tests, because as far as the test infrastructure is concerned, they're both just strings, but I guess it doesn't hurt to have a test with a short string and a long string just in case of some extremely obscure error that would affect one case and not the other, and having both cases in the tests certainly does give them more value as "documentation".

And yeah, things will look cleaner after PEP 639 but we can deal with that as it comes up.

diazona:
What's the timeline looking like for this? I think it's probably going to be the last thing blocking an initial release (or at least, among the things I added to the "initial release" milestone)

sjlongland:
Yeah, this has taken way longer than I wanted.  Largely because I've been busy with work commitments before I went on leave.  I'm now officially on leave until 2023-11-06 so I have some time now to clean things up.

tests/test_license_files.py
sjlongland: Realistically, both… if the file exists, we shouldn't be overwriting it should we?

Or do we just overwrite it regardless?

tests/test_license_files.py
sjlongland: Yep, I'll fix that up now… just got to look up what the syntax is for RST. (I'm too used to Markdown, sadly the Markdown situation with Sphinx is sketchy so RST really is the better choice)

tests/test_license_files.py
sjlongland: Actually, just thinking about this… `setuptools` would have _expected_ the files to already exist if they were listed in `setup.cfg`.

We're simply moving the configuration setting from `setup.cfg` to its new place in `pyproject.toml`, so the files _should_ already be there.  Maybe we should be raising a warning if the file is not found at the time of migration?  Otherwise, `setuptools` should suitably squawk.

Thinking about it now, `README` was a little different -- you could specify the README text inline in the `long_description`, and the code we have migrates this to a separate file and links it in `pyproject.toml`.

diazona: Whoops, sorry I missed these comments!

I think our philosophy has generally been to stay as close as possible to a "dumb" translation of `setup.cfg` fields into `setup.py` fields. We generally rely on setuptools to validate things as much as possible, because like you said, it _is_ going to check field values as needed when it builds the project. With that in mind, I'd say it makes sense to just copy the file names from `setup.cfg` to `pyproject.toml` without checking whether they exist, or definitely without _failing_ if they don't exist. An mild warning wouldn't be out of place, but personally I would be totally fine just skipping it as well.

I can imagine situations where the file might not exist and it'd be fully legitimate - like that web-based converter we were talking about once upon a time, where people might want to only upload `setup.cfg` and/or `setup.py` but not have to add in other random files that don't actually need to be read during the conversion.

tests/test_license_files.py
diazona: Haha yeah I can relate

tests/test_license_files.py
diazona: ...as far as what to actually do here: at minimum I'd suggest rephrasing the docstring. If the intent is to contrast with the other test, which uses globs, and point out that this one does not use globs, perhaps something like this:

```suggestion
    Test that concrete file names are added as paths.
```

Or maybe there's even a better term than "concrete", I'm not sure :shrug: but hopefully something along those lines will be clear enough.

diazona (changes requested):

diazona:
LGTM now! Did you want to squash/reorganize the commits at all before this gets merged? (Just figured I'd ask because, you know, auto-merge is enabled and if I just click "approve" then it'll be too late to do any commit reorganization)

sjlongland:
> LGTM now! Did you want to squash/reorganize the commits at all before this gets merged? (Just figured I'd ask because, you know, auto-merge is enabled and if I just click "approve" then it'll be too late to do any commit reorganization)

I can squash it if you feel that'd be better… to be honest I find having the full commit history is useful when troubleshooting because tools like `git bisect` then has much finer-grained changes to work with.

If you bundle multiple changes into a single commit, `git bisect` starts to use its usefulness due to the coarseness of commits.

diazona:
Oh for sure, I definitely agree in keeping things broken out for bisection and similar use cases. Sorry for the confusion: I didn't mean squashing everything together into a single commit, but I was thinking more like squashing some of the little commits that don't really stand on their own, like typo fixes or whatever, into the more standalone commits, just to reduce clutter in the commit history and eliminate intermediate steps that don't compile. In other words, I'm talking about an interactive rebase, not a squash-merge (which would be combining all the commits together into one). This is something I do routinely (although usually before pushing) so I do think it's a good idea, but it's not something import enough that I would try to insist on it. Totally up to you whether you want to do this or not.

I will be happy to approve either way, I'm just making sure I don't take the option away.

sjlongland:
Okay, did a final local test after a slight flattening of the commit history:

```
===================================== 107 passed, 4 xfailed, 2 warnings in 1980.75s (0:33:00) =====================================
.pkg: _exit> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta
  py: OK (2154.31=setup[172.05]+cmd[1982.26] seconds)
  congratulations :) (2154.72 seconds)
```

… took a bit longer than I was expecting.  Hopefully CI won't take this long.

diazona:
Whoa, that is a looong time.

It might be related to the distribution package testing, especially if it's the first time you're running those tests - it would have to download the source code for pytest and pytest-localserver. But still I wouldn't expect it to be that long (33 minutes!) unless you have a very slow internet connection or something like that. In any case, it looks like pre-merge tests in CI were still relatively quick in comparison.

diazona (approved):

sjlongland:
Yep, I think it is Internet connection related… 25Mbps down / 2Mbps up hybrid-fibre coax.

The laptop here is no spring chicken, but it's not _that_ slow.  `ps axf` showed a lot of time spent running `pip`.

PR #90, opened 2023-10-05
Enable towncrier for changelog aggregation

### LINKED ISSUE #88: Enable towncrier
We have a configuration file for [towncrier](https://towncrier.readthedocs.io/en/stable/) in the repository (inherited from https://github.com/jaraco/skeleton), so I figure we might as well make use of it. All the cool projects have a changelog, and using towncrier to maintain one is going to be slightly less annoying than doing it manually.

Actually I started working on this a little already a couple days ago while I was waiting for some other stuff, just to see how tricky it would be... seems like it's fairly straightforward. The only significant change is that every PR now has to include a changelog fragment.

@sjlongland any thoughts on this? I think if we're going to do it, best to get it out of the way now. (This should be my last last-minute addition to the initial release milestone :stuck_out_tongue:)

diazona:
This PR sets up the configuration and infrastructure for [towncrier](https://towncrier.readthedocs.io/en/stable/index.html) to maintain our changelog. Going forward, every pull request should contain a changelog fragment. We can add something to verify that, but I've decided to omit that from the scope of this PR because I don't want to stretch it out too long.

With this change, our procedure for releasing the package (previously outlined in #40) gains an extra step:

1. After everything else needed for the release has been done, locally create a branch and run `tox run -e towncrier -- version {version}` (where `{version}` is a placeholder for the actual version number we want to release). That will generate the changelog entries. Commit the changes, push the branch, make a PR, and get it merged by the usual process.
2. Then make a Github release (referring to the announcement page on Github that we make using [this form](https://github.com/diazona/setuptools-pyproject-migration/releases/new)), with a corresponding tag `v{version}`, in the Github web interface (or by using the Github API) - this is step 1 from the first list in the description of #40.
3. Continue with the rest of the procedure as laid out in that list.

It's going to be a little tedious at first, but I think this is not too bad. In the future, we can definitely make this more automatic e.g. by adding a Github workflow to run through some of these steps, but it may be better to switch around some of the steps the release procedure, perhaps along the lines of one of the alternatives described in #40. I think we need some time (and maybe experience) to figure that out, so I've chosen to defer doing anything more complicated for a future PR. (Open to discussion, as always)

Closes #88 (and take note of the discussion happening there before deciding whether to merge this!)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,6 +49,11 @@ repos:
   hooks:
   - id: markdownlint-fix
 
+- repo: https://github.com/twisted/towncrier
+  rev: 23.6.0
+  hooks:
+  - id: towncrier-check
+    files: $changelog\.d/
 ci:
   skip:
   # pre-commit.ci doesn't have Docker available
--- a/changelog.d/88.feature
+++ b/changelog.d/88.feature
@@ -1,0 +1,1 @@
+Enable towncrier
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -17,7 +17,7 @@ html_theme = "furo"
 # Link dates and other references in the changelog
 extensions += ["rst.linker"]
 link_files = {
-    "../NEWS.rst": dict(
+    "../CHANGELOG.rst": dict(
         using=dict(GH="https://github.com"),
         replace=[
             dict(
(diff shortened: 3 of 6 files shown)


diazona:
P.S. I am very tired at the moment and could easily have made a few silly mistakes, so a nice careful review of this code would be especially appreciated :smile:

sjlongland (approved):

sjlongland (approved):

PR #96, opened 2023-10-08
Set long description content type to text/markdown

### LINKED ISSUE #95: Project description does not display properly on PyPI
Looking at the [PyPI page for our initial prerelease](https://pypi.org/project/setuptools-pyproject-migration/0.1a0/), the description is all messed up - it displays in raw Markdown rather than being properly formatted. That needs to be fixed.

diazona:
Prior to this change, the [description on PyPI](https://pypi.org/project/setuptools-pyproject-migration/) shows up as raw Markdown code. Hopefully this will get PyPI to interpret it as Markdown and show the properly formatted version.

Closes #95

### DIFF
--- a/changelog.d/99.bugfix.txt
+++ b/changelog.d/99.bugfix.txt
@@ -1,0 +1,3 @@
+
+Set the long description content type to Markdown in order to
+(hopefully) make the description render properly on PyPI.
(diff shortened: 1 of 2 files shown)


sjlongland (approved):
That's the nice thing about standards, there's so many to choose from. :-)

Issue #100, opened 2023-10-10
ModuleNotFoundError: No module named 'packaging'


nedbat:
I'm not sure what's different about my environment:

```
% pipx install setuptools-pyproject-migration
  installed package setuptools-pyproject-migration 0.1.0, installed using Python 3.11.5
  These apps are now globally available
    - setup-to-pyproject
done! ✨ 🌟 ✨

% setup-to-pyproject
Traceback (most recent call last):
  File "/Users/nbatchelder/.local/bin/setup-to-pyproject", line 5, in <module>
    from setuptools_pyproject_migration.cli import main
  File "/Users/nbatchelder/.local/pipx/venvs/setuptools-pyproject-migration/lib/python3.11/site-packages/setuptools_pyproject_migration/__init__.py", line 7, in <module>
    from packaging.specifiers import SpecifierSet
ModuleNotFoundError: No module named 'packaging'
```

diazona:
Ooh, good catch, thanks for reporting! (congrats on the first external issue report 🎉) We probably just forgot to list a dependency. We'll definitely fix it - or feel free to submit a PR if you want, no obligation though.

Would you mind sharing the exact list of packages that got installed for you? (`pipx runpip setuptools-pyproject-migration freeze`) Just in case it's relevant for version info or something.

nedbat:
```
% pipx runpip setuptools-pyproject-migration freeze
Parsley==1.3
pep508-parser==2019.3
setuptools-pyproject-migration==0.1.0
tomlkit==0.12.1
typing_extensions==4.8.0
```

sjlongland:
On Gentoo, the missing module is provided by `dev-python/packaging:0` … homepage: https://pypi.org/project/packaging/

`qdepends` tells me `dev-python/setuptools` is one of the packages that depends on it.  `setuptools` is explicitly listed as a dependency, so not quite sure how that got missed.

sjlongland:
Can reproduce:

```
RC=0 stuartl@rikishi ~/projects/wicen/rfid/aioax25 $ setup-to-pyproject 
Traceback (most recent call last):
  File "/home/stuartl/.local/bin/setup-to-pyproject", line 5, in <module>
    from setuptools_pyproject_migration.cli import main
  File "/home/stuartl/.local/pipx/venvs/setuptools-pyproject-migration/lib/python3.11/site-packages/setuptools_pyproject_migration/__init__.py", line 7, in <module>
[...]

PR #101, opened 2023-10-10
Add missing `packaging` dependency explicitly

sjlongland:
We were relying on `setuptools` implicitly pulling it in, but it seems that does not happen when `pipx` is used to install everything.  So instead, explicitly list it.

Closes issue #100.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,6 +8,13 @@ usedevelop = True
 extras =
 	testing
 
+[testenv:script]
+# Deliberately blank to override the defaults set above
+deps =
+extras =
+commands =
+	setup-to-pyproject --help
+
 # There is very little advantage to using the pre_commit tox environment,
 # compared to running pre-commit directly, but we provide it in case you want to
 # use it. For example, if you don't want to install pre-commit locally.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,6 +69,9 @@ jobs:
           allow-prereleases: true
       - name: Install tox
         run: python -m pip install tox
+      - name: Run script checks
+        run: >
+          tox -e script
       - name: Run tests
         # For simplicity, we limit forced minimum dependencies to direct
         # dependencies and build system dependencies, not extra dependencies
--- a/changelog.d/100.bugfix.rst
+++ b/changelog.d/100.bugfix.rst
@@ -1,0 +1,2 @@
+Explicitly list ``packaging`` as a dependency so ``pipx`` installs it in the virtual environment instead of relying on
+``setuptools`` to pull it in.
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,6 +28,7 @@ package_dir =
 include_package_data = true
 python_requires = >=3.6
 install_requires =
+	packaging
 	pep508-parser
 	setuptools
 	tomlkit
@@ -53,7 +54,6 @@ testing =
 	# local
 	importlib-metadata; \
 		python_version < "3.8"
-	packaging
 	# no version of pyproject-metadata supports Python 3.6
 	pyproject-metadata; \
 		python_version >= "3.7"


diazona:
Is there a way we can set up an automated test to reproduce the bug? I mean, I would have expected `tox` to cover this because it provides isolated environments, but I guess `packaging` gets installed in those environments anyway for some reason...?

sjlongland:
> Is there a way we can set up an automated test to catch the bug? I mean, I would have expected `tox` to cover this because it provides isolated environments, but I guess `packaging` gets installed in those environments anyway for some reason...?

Not certain on that, but the fact we've now explicitly listed it, means if it _doesn't_ get installed, that's a `pip` bug.  I think we'd have a hard time convincing `tox` to not install the package to trigger the issue.

diazona:
Yeah that's true, sorry; what I mean is I want to include the test that should have caught this bug originally. I'm not saying we should try to fake the absence of the `packaging` package.

The goal here is to catch other similar errors that may exist or may be introduced in the future. Maybe not `packaging` itself, but missing dependencies that might occasionally be pulled in by other packages if we're not careful. (Though there is the slight chance that a future typo deletes the dependency on `packaging`)

sjlongland:
Ahh right, so future dependency whoopsies…

https://www.tweag.io/blog/2023-03-14-announcing-fawltydeps/ might be able to help, I'm doing some experiments with it now.

sjlongland:
`fawltydeps` _almost_ gets there.  I'm testing on the present `main` branch (tag `v0.1.0`)

The big problem is that `setup.cfg` does not list `packaging` in `install_requires`, but _does_ list it for the optional `testing` package environment.  `fawltydeps` sees this, then declares that indeed, the dependency is resolved.

I can't see a way for it to "ignore" these "extras", it just assumes we're installing with all the extras.

https://github.com/tweag/FawltyDeps/issues/44#issuecomment-1377003779 might resolve that for us.

diazona:
Seems handy! That could be a useful addition to our test suite. But personally I'd be happiest having a test that actually runs our console script and would fail with that `ImportError` if `packaging` is not installed. (TBH a big part of the reason I'd like that is that it makes me uncomfortable bot knowing why our existing tests didn't catch that problem.)

sjlongland:
This is getting closer:

```
RC=0 stuartl@rikishi ~/projects/setuptools-pyproject-migration/work $ tox -e check_depends
check_depends: recreate env because python changed virtualenv version='20.24.4'->'20.24.2'
check_depends: remove tox env folder /home/stuartl/projects/setuptools-pyproject-migration/work/.tox/check_depends
.pkg: recreate env because python changed virtualenv version='20.24.4'->'20.24.2'
.pkg: remove tox env folder /home/stuartl/projects/setuptools-pyproject-migration/work/.tox/.pkg
.pkg: install_requires> python -I -m pip install setuptools 'setuptools>=56' 'setuptools_scm[toml]>=3.4.1'
.pkg: _optional_hooks> python /home/stuartl/projects/setuptools-pyproject-migration/env/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: get_requires_for_build_editable> python /home/stuartl/projects/setuptools-pyproject-migration/env/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: install_requires_for_build_editable> python -I -m pip install wheel
.pkg: build_editable> python /home/stuartl/projects/setuptools-pyproject-migration/env/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta
check_depends: install_package_deps> python -I -m pip install pep508-parser setuptools tomlkit 'typing-extensions<5,>=4'
check_depends: install_package> python -I -m pip install --force-reinstall --no-deps /home/stuartl/projects/setuptools-pyproject-migration/work/.tox/.tmp/package/94/setuptools_pyproject_migration-0.1.1.dev0+gfde5f75.d20231011-0.editable-py3-none-any.whl
check_depends: commands[0]> setup-to-pyproject --help
  File "/home/stuartl/projects/setuptools-pyproject-migration/work/.tox/check_depends/bin/setup-to-pyproject", line 5, in <module>
    from setuptools_pyproject_migration.cli import main
  File "/home/stuartl/projects/setuptools-pyproject-migration/work/src/setuptools_pyproject_migration/__init__.py", line 7, in <module>
    from packaging.specifiers import SpecifierSet
ModuleNotFoundError: No module named 'packaging'
check_depends: exit 1 (0.36 seconds) /home/stuartl/projects/setuptools-pyproject-migration/work> setup-to-pyproject --help pid=242725
.pkg: _exit> python /home/stuartl/projects/setuptools-pyproject-migration/env/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta
  check_depends: FAIL code 1 (8.65=setup[8.29]+cmd[0.36] seconds)
  evaluation failed :( (8.85 seconds)
```

```
diff --git a/tox.ini b/tox.ini
index 2b63e2a..1573ea9 100644
--- a/tox.ini
+++ b/tox.ini
@@ -8,6 +8,15 @@ usedevelop = True
 extras =
        testing
 
+# Check whether there's dependencies present to even make the script run.
+[testenv:check_depends]
+deps =
+extras =
+setenv =
+       PYTHONWARNDEFAULTENCODING = 1
+commands =
+       setup-to-pyproject --help
+
 # There is very little advantage to using the pre_commit tox environment,
 # compared to running pre-commit directly, but we provide it in case you want to
 # use it. For example, if you don't want to install pre-commit locally.
```

sjlongland:
```
pyproject-migration/setuptools-pyproject-migration/.tox/.tmp/package/1/setuptools_pyproject_migration-0.1.dev1%2Bg4d2b46b-0.editable-py3-none-any.whl#sha256=b73ff7f877782a998096c504202b50af9de33bd7ae32db4afc8209abdd1e3fbe,tomlkit==0.12.1,typing_extensions==4.8.0,wheel==0.41.2[0m
2023-10-11T01:54:50.2336611Z [1m[35mcheck_depends:[0m[36m commands[0][22m[2m>[0m setup-to-pyproject --help[0m
2023-10-11T01:54:50.4795481Z usage: setup-to-pyproject [-h]
2023-10-11T01:54:50.4795720Z 
2023-10-11T01:54:50.4795977Z Run the pyproject setuptools command. This effectively does the same thing as
2023-10-11T01:54:50.4796282Z 
2023-10-11T01:54:50.4796809Z     python setup.py pyproject
2023-10-11T01:54:50.4797017Z 
2023-10-11T01:54:50.4797386Z except that if setup.py doesn't exist, it will act as though there were a "stub"
2023-10-11T01:54:50.4797816Z setup.py script with the following contents:
2023-10-11T01:54:50.4798038Z 
2023-10-11T01:54:50.4798152Z     import setuptools
2023-10-11T01:54:50.4798433Z     setuptools.setup()
2023-10-11T01:54:50.4798631Z 
2023-10-11T01:54:50.4798990Z Effectively, this lets you use setuptools-pyproject-migration without having to
2023-10-11T01:54:50.4799603Z install the plugin in your project's virtual environment, and without having to
2023-10-11T01:54:50.4800056Z create a setup.py file if all you have is setup.cfg.
2023-10-11T01:54:50.4800287Z 
2023-10-11T01:54:50.4800378Z options:
2023-10-11T01:54:50.4800728Z   -h, --help  show this help message and exit
2023-10-11T01:54:50.5064150Z [1m[35m.pkg:[0m[36m _exit[22m[2m>[0m python /opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta[0m
2023-10-11T01:54:50.5498446Z   check_depends: OK (7.00=setup[6.73]+cmd[0.27] seconds)
2023-10-11T01:54:50.5499169Z   congratulations :) (7.13 seconds)
2023-10-11T01:54:50.5925201Z Post job cleanup.
2023-10-11T01:54:50.7542158Z Post job cleanup.
2023-10-11T01:54:50.8594846Z [command]/usr/bin/git version
2023-10-11T01:54:50.8646491Z git version 2.42.0
2023-10-11T01:54:50.8693164Z Temporarily overriding HOME='/home/runner/work/_temp/95da1d46-8c78-4b24-b052-b3dc26d3cfd7' before making global git config changes
2023-10-11T01:54:50.8694895Z Adding repository directory to the temporary git global config as a safe directory
```

Looks like CI is running it.

tox.ini
```diff
@@ -8,6 +8,15 @@ usedevelop = True
 extras =
 	testing
 
+# Check whether there's dependencies present to even make the script run.
+[testenv:check_depends]
+deps =
+extras =
```
diazona, line 14: No need for these if we're not giving them values

```suggestion
```

tox.ini
```diff
@@ -8,6 +8,15 @@ usedevelop = True
 extras =
 	testing
 
+# Check whether there's dependencies present to even make the script run.
+[testenv:check_depends]
```
diazona, line 12: Personal opinion: I would give this a more generic name because it might be useful for more than just checking dependencies at some point.

```suggestion
[testenv:script]
```

or it could be some other name, but the idea is that this environment just runs the script, so we name it after that, rather than naming it after whatever assorted things we expect it to check.

.github/workflows/main.yml
```diff
@@ -56,6 +56,25 @@ jobs:
...
+  check_depends:
+    name: Run dependency checks
+    uses: ./.github/workflows/check_depends.yml
+    # https://wildwolf.name/github-actions-how-to-avoid-running-the-same-workflow-multiple-times/
+    if: >
+      github.event_name != 'pull_request'
+      || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
+    strategy:
+      matrix:
+        # We run on just one release of Python and one OS, as they should be all alike.
```
diazona, line 68: Building off my other suggestion of giving the tox environment a more generic name, I would actually suggest just running this check from the normal `tests.yml` workflow. It might be useful for more than just ensuring all the dependencies are installed, and whatever other stuff it does check could conceivably vary from one Python version to another. (And actually, dependencies can be constrained to specific Python versions so there's probably some value to running the check on all Python versions just because of that.)

So I'd suggest just putting a separate step in `tests.yml` with a separate `tox` invocation that uses the new environment.

changelog.d/100.fixed
```diff
@@ -0,0 +1,2 @@
+Explicitly list `packaging` as a dependency so `pipx` installs it in the virtual environment instead of relying on
+`setuptools` to pull it in.
```
diazona, line 2: According to [the documentation](https://towncrier.readthedocs.io/en/stable/tutorial.html#creating-news-fragments) it's `bugfix`, not `fixed`. (unless it accepts both? But still probably best to be standard)

Also I think you might be able to give this a `.md` suffix to have it interpreted as Markdown. Like `100.bugfix.md`.

diazona (changes requested):

tox.ini
sjlongland: Ahh okay, I thought setting them empty was overriding the defaults set in the top-level `testenv` section of `tox.ini`.  I saw `deps =` there already, so thought it had that specific meaning.

changelog.d/100.fixed
sjlongland: Yep, no problems… I thought it was just `fixed`, but I'll rename. :-)

tox.ini
diazona: ohhh wait I think you're right, sorry!

Though I'd say technically the proper way to fix that is to make the top-level section header into `[testenv:py,py3{6,7,8,9,10,11,12}]` so that it explicitly lists environments, rather than defining defaults. But if you don't feel like getting into that as part of this PR, no worries.

diazona (changes requested):
One last change I'd suggest - not the most essential thing but I think it'd be an improvement.

changelog.d/100.bugfix
```diff
@@ -0,0 +1 @@
+Explicitly list ``packaging`` as a dependency so ``pipx`` installs it in the virtual environment instead of relying on ``setuptools`` to pull it in.
```
diazona, line 1: wait did it lose my comment? :facepalm:

I was just suggesting renaming this to `100.bugfix.rst`; I think it's a good idea in general to make sure these files have the right extension for their formatting. Sorry if you see this twice.

diazona (approved):

Issue #102, opened 2023-10-11
setuptools-pyproject-migration barfs if `long_description` specified in `setup.py`


sjlongland:
So, `aioax25` actually specifies everything in `setup.py`, including reading the `README.md` itself and importing the to-be-installed module to pick up the version number.  Supporting that kind of abuse is probably against the scope of this project, but thankfully, the hand-edits to make it work weren't big:

I hand-edited the `setup.py` from [this](https://github.com/sjlongland/aioax25/blob/a963330b658b6d2934e223853b64f3721c4d6ec8/setup.py) to this:
```python
#!/usr/bin/env python3

from setuptools import setup
from os.path import dirname, join
from sys import version_info

requirements = [
        'pyserial',
        'signalslot',
        'pyserial_asyncio'
]

packages = [
        'aioax25',
        'aioax25.aprs',
]

setup(
        name='aioax25',
        url='https://github.com/sjlongland/aioax25/',
        version="0.0.12",
        author='Stuart Longland VK4MSL',
        author_email='me@vk4msl.id.au',
        license='GPL-2.0-or-later',
        packages=packages,
        requires=requirements,
        install_requires=requirements,
        description='Asynchronous AX.25 interface in pure Python using asyncio',
        long_description='file:README.md',
        long_description_content_type='text/markdown',
        classifiers=[
            'Development Status :: 2 - Pre-Alpha',
            'Environment :: No Input/Output (Daemon)',
            'Framework :: AsyncIO',
            'Intended Audience :: Developers',
            'License :: OSI Approved :: GNU General Public License v2 or later (GPLv2+)',
            'Operating System :: POSIX',
            'Programming Language :: Python :: 3 :: Only',
            'Topic :: Communications :: Ham Radio'
        ]
)
```

This crashes, because when `setuptools` is invoked in this manner, it does _not_ create a `metadata` key in the `command_options` property `dict`:

```
  File "/home/stuartl/.local/pipx/shared/lib/python3.11/site-packages/setuptools/_distutils/core.py", line 201, in run_commands
    dist.run_commands()
  File "/home/stuartl/.local/pipx/shared/lib/python3.11/site-packages/setuptools/_distutils/dist.py", line 969, in run_commands
    self.run_command(cmd)
  File "/home/stuartl/.local/pipx/shared/lib/python3.11/site-packages/setuptools/dist.py", line 989, in run_command
    super().run_command(command)
[...]

PR #103, opened 2023-10-11
Handle README being specified in setup.py

sjlongland:
Annoyingly, `setuptools` omits the `metadata` property of `command_options`, and this causes the `pyproject.toml` generator to barf rather spectacularly.

So, check for the existence of this key, and if not present, we fail back to using the getter methods and properties of the `Distribution` to retrieve this information.

Fixes issue #102.

### DIFF
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -278,12 +278,10 @@ class WritePyproject(setuptools.Command):
             pyproject["project"]["description"] = description
 
         if dist.get_long_description() not in (None, "UNKNOWN"):
+            long_description_source: str = dist.get_long_description()
+
             if "metadata" in dist.command_options:
-                long_description_source: str = dist.command_options["metadata"]["long_description"][1]
-            else:
-                # setuptools does not tell us where this text came from, so we put it inline so it is at least
-                # preserved!
-                long_description_source: str = dist.get_long_description()
+                long_description_source = dist.command_options["metadata"]["long_description"][1]
 
             long_description_content_type: Optional[str] = dist.metadata.long_description_content_type
 
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -278,10 +278,12 @@ class WritePyproject(setuptools.Command):
             pyproject["project"]["description"] = description
 
         if dist.get_long_description() not in (None, "UNKNOWN"):
-            long_description_source: str = dist.get_long_description()
+            long_description_source: str
 
             if "metadata" in dist.command_options:
                 long_description_source = dist.command_options["metadata"]["long_description"][1]
+            else:
+                long_description_source = dist.get_long_description()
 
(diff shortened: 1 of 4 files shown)


src/setuptools_pyproject_migration/__init__.py
```diff
@@ -278,8 +278,13 @@ def _generate(self) -> Pyproject:
             pyproject["project"]["description"] = description
 
         if dist.get_long_description() not in (None, "UNKNOWN"):
-            long_description_source: str = dist.command_options["metadata"]["long_description"][1]
+            long_description_source: str = dist.get_long_description()
+
+            if "metadata" in dist.command_options:
+                long_description_source = dist.command_options["metadata"]["long_description"][1]
```
diazona, line 284: Personally I'd write it like this, mostly because it looks cleaner, but this also avoids one extra call to `get_long_description()` under certain circumstances:

```suggestion
            long_description_source: str
            if "metadata" in dist.command_options:
                long_description_source = dist.command_options["metadata"]["long_description"][1]
            else:
                long_description_source = dist.get_long_description()
```

Up to you though, I'd call this one an opinion.

diazona (approved):
A couple other things that come to mind:

- There's probably some opportunity to reduce duplication in the tests, maybe with a parametrized fixture that can run a test using either `setup.cfg` or `setup.py` to store the metadata. But that's certainly not a requirement for this PR. (or, really, ever; it'd just be a minor improvement as far as I'm concerned)
- Since `aioax25` motivated this change, I think it'd be a good opportunity to add that as a distribution package test. But again, it definitely doesn't have to be done as part of this PR.

src/setuptools_pyproject_migration/__init__.py
sjlongland: Ahh okay, I wasn't sure how to declare the variable type without assigning it a value.  Before, it moaned at me because I was declaring the type in two places.

sjlongland:
> A couple other things that come to mind:
> 
>     * There's probably some opportunity to reduce duplication in the tests, maybe with a parametrized fixture that can run a test using either `setup.cfg` or `setup.py` to store the metadata. But that's certainly not a requirement for this PR. (or, really, ever; it'd just be a minor improvement as far as I'm concerned)

Yep, I wasn't sure how to best tackle this, but you're right, there's definitely some duplication going on that we could avoid.  Might be a good follow-up issue.

>     * Since `aioax25` motivated this change, I think it'd be a good opportunity to add that as a distribution package test. But again, it definitely doesn't have to be done as part of this PR.

We could do that, you'll want to use release `0.0.12` as the benchmark, as I've just merged a commit to the `master` branch that now uses `pyproject.toml` and replaces `setup.py` with a very simple script.

src/setuptools_pyproject_migration/__init__.py
diazona: Yeah at some point in the past I ran into the same problem and had to look up how to do it. It's not the most obvious thing :sweat_smile:

diazona (approved):

PR #106, opened 2023-10-12
Use full history and ignore local changes in ReadTheDocs

### LINKED ISSUE #99: Version number does not display properly on ReadTheDocs
In the [documentation for our stable release 0.1](https://setuptools-pyproject-migration.readthedocs.io/en/stable/), the version number at the top left is displayed as "setuptools-pyproject-migration 0.1.1.dev0+gfde5f75.d20231010". We need to fix that so it displays properly, e.g. for this release it should show `0.1`, or for future releases something like `0.1.post0` or `0.1.1` or `0.2` or whatever.

diazona:
When ReadTheDocs builds our documentation, it runs into a couple of problems determining the version number:

- It uses a shallow clone by default, which won't have the information about the latest Git tag which `setuptools_scm` uses to determine the version number.

    The solution recommended [in ReadTheDocs' documentation](https://docs.readthedocs.io/en/stable/build-customization.html#unshallow-git-clone) is to unshallow the clone by fetching the full history as a post-checkout step.
- It makes some local changes to files including `docs/conf.py`. `setuptools_scm` then checks the status of the Git working tree when determining the package version, and comes up with a version number suitable for a local development version, like `0.1.1.dev16+g124dc35.d20231012`, which then appears in the generated documentation. But we want a final release version number like `0.1.1`, at least for versions of the documentation generated from final release tags.

    The solution recommended [in ReadTheDocs' documentation](https://docs.readthedocs.io/en/stable/build-customization.html#avoid-having-a-dirty-git-index) is to set the assume-unchanged bit on the modified working tree files so Git doesn't treat them as modified.

This PR implements these two changes. Hopefully this is enough to make the version number look correct.

Closes #99 (hopefully)

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -11,5 +11,7 @@ build:
   tools:
     python: "3"
   jobs:
+    post_checkout:
+    - '[ "$(git rev-parse --is-shallow-repository)" == "true" ] && git fetch --unshallow --tags'
     pre_install:
     - git update-index --assume-unchanged docs/conf.py
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -10,3 +10,6 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3"
+  jobs:
+    pre_install:
+    - git update-index --assume-unchanged docs/conf.py
--- a/changelog.d/99.bugfix.txt
+++ b/changelog.d/99.bugfix.txt
@@ -1,0 +1,2 @@
+When building documentation on ReadTheDocs, use a full (non-shallow) clone and
+ignore local modifications, in order to get the right version number


diazona:
I'm running a quick test by creating a dummy tag that looks like a version number, `v0.1.0.1`. I'll delete the tag (and hide the generated version of the documentation) afterwards.

diazona:
OK so this is interesting: ReadTheDocs generated [a version of the documentation from the tag itself](https://setuptools-pyproject-migration.readthedocs.io/en/v0.1.0.1/), which comes out correct, and also generated a version from the same _commit_ but a detached head which came out with the wrong version number. So I think we need to either tell it to fetch tags before building documentation, or set up some kind of rule on ReadTheDocs to specifically use the tag versions.

Incidentally I kind of screwed up the `stable` build of the documentation by pushing a tag, and I can't seem to fix it... but I think it will take care of itself when we release another proper version, which hopefully won't be too long. And ReadTheDocs uses `latest` by default instead of `stable` (at least until we configure it otherwise), so it probably won't hurt to just let it go for now.

sjlongland (approved):

sjlongland (approved):

PR #111, opened 2023-10-24
Strip README path

adamchainz:
setuptools allows whitespace after `file:`, as in `file: README.rst`. It needs stripping when converting to TOML, otherwise setuptools will fail to find the README.

### DIFF
--- a/tests/test_readme.py
+++ b/tests/test_readme.py
@@ -110,6 +110,21 @@ setuptools.setup(
     assert readme_path.read_text(encoding="utf-8").rstrip("\n") == long_description
 
 
+def test_file_with_space(project) -> None:
+    setup_cfg = """\
+[metadata]
+name = test-project
+version = 0.0.1
+long_description = file: README.rst
+"""
+
+    project.setup_cfg(setup_cfg)
+    project.write("README.rst", "Dummy README\n")
+    project.setup_py()
+    result = project.generate()
+    assert result["project"]["readme"] == "README.rst"
+
+
 @parametrize_readme_type
 def test_file_with_content_type(project, extension: str, mime_type: str) -> None:
     readme_filename = f"README.{extension}"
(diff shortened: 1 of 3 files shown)


diazona:
Thanks! Could you add a test that verifies this behavior? (or someone else can help with that if you don't want to do it) 

Also don't forget to add a changelog fragment using towncrier. The [towncrier docs](https://towncrier.readthedocs.io/en/stable/cli.html#towncrier-create) should explain how - basically run `towncrier create` with suitable options - but let me know if you want more specific instructions.

adamchainz:
Done and done!

tests/test_readme.py
```diff
@@ -110,6 +110,21 @@ def test_string_without_content_type_setuppy(project) -> None:
     assert readme_path.read_text(encoding="utf-8").rstrip("\n") == long_description
 
 
+def test_file_with_space(project) -> None:
+    setup_cfg = """\
+[metadata]
+name = test-project
+version = 0.0.1
+long_description = file: README.rst
```
sjlongland, line 118: Do we know if `setup.cfg` quietly strips trailing whitespace from strings?  I know it's an INI-file derivative, but not 100% certain on the semantics.

I'm confident the code will handle it anyway, so not overly concerned, but it might be helpful to add a similar test that makes sure.

diazona:
Thanks! One last thing, sorry: the commit message is rather sparse as is, so would you mind adding putting a bit more detail about what the bug was and what your fix is, and mention that you added a test for it?

tests/test_readme.py
diazona: Oh, yeah good idea, although personally I wouldn't consider it a blocker for this PR. I just wanted to make sure there is _some_ test exercising the change, but additional tests for edge cases can always come later.

adamchainz:
How’s this now?

diazona (approved):
Works for me 👌

tests/test_readme.py
diazona: Documented in #115, and we can figure out later what makes sense.

PR #114, opened 2023-10-25
Improve inline array and table creation

adamchainz:
Follow-up to #112 with some changes I was working on after the comments there.

1. Add a changelog note.
2. Make a generic function for making inline arrays and tables, to unclutter `_generate()`.
3. Don't turn `urls` into an inline table, as it is too cramped.
4. Add recursive assertion on types in `test_setup_command.py`.

### DIFF
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -6,7 +6,8 @@ import tomlkit
 import warnings
 from packaging.specifiers import SpecifierSet
 from pep508_parser import parser as pep508
-from typing import Dict, List, Optional, Set, Tuple, Type, Union
+from tomlkit.api import Array, InlineTable
+from typing import Dict, List, Optional, Set, Tuple, Type, TypeVar, Union
 
 # After we drop support for Python <3.10, we can import TypeAlias directly from typing
 from typing_extensions import Required, TypedDict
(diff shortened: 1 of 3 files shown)


src/setuptools_pyproject_migration/__init__.py
```diff
@@ -282,9 +288,7 @@ def _generate(self) -> Pyproject:
 
         urls: Dict[str, str] = dist.metadata.project_urls
         if urls:
-            urls_table = tomlkit.inline_table()
-            urls_table.update(urls)
-            pyproject["project"]["urls"] = urls_table
+            pyproject["project"]["urls"] = urls
```
sjlongland, line 291: Did we intend to revert the `urls` field here from being an inline table back to a regular table?

sjlongland (commented):
Changes look good, just a little query on one change, we seem to be reverting to what was there before.

src/setuptools_pyproject_migration/__init__.py
adamchainz: Yes, since inline tables are one line they get cramped with URLs:

```toml
urls = {Homepage = "https://example.com/repo/example", Docs = "https://docs.example.com", Funding = "https://example.com/funding/", etc = "and so on and so forth"}
```

sjlongland (approved):

PR #123, opened 2023-12-04
Add a distribution package test for aioax25

### LINKED ISSUE #105: Add aioax25 0.0.12 as a distribution package test
This package exposes the issue where the long description is specified only in `setup.py` and there is no `setup.cfg`. (#102) We should add a distribution package test to check that the plugin can handle that.

Based on [a comment on PR #103](https://github.com/diazona/setuptools-pyproject-migration/pull/103#pullrequestreview-1669762001)

diazona:
This is the long-awaited (by me, probably not anyone else) implementation of a test of [aioax25](https://pypi.org/project/aioax25/), to confirm that setuptools-pyproject-migration produces a `pyproject.toml` file that generates the same core metadata as the project's own `setup.py` file. In order to get this working, I kind of had to rewrite the distribution package test support code. Details are in the commit messages, but basically:

- I added support for testing packages that only provide sdists
- I added support for testing packages that require their own source code to be importable when running `setup.py`
- I refactored the `DistributionPackage`/`DistributionPackagePreparation` class hierarchies so that `DistributionPackage` would not have any state that needs to be changed during testing
- I also changed the distribution package test function to a test class, with individual methods for each of the fields, to give better visibility into which fields are and aren't getting properly computed and better error reporting

Fortunately, after all that work, the new test passes :smile: 

Closes #105

### DIFF
--- a/changelog.d/105.feature.rst
+++ b/changelog.d/105.feature.rst
@@ -1,0 +1,1 @@
+Support testing against live distribution packages that don't provide wheels by computing the metadata from the sdist, and against packages that require their own source code to be importable from ``setup.py``
--- a/changelog.d/105.misc.rst
+++ b/changelog.d/105.misc.rst
@@ -1,0 +1,19 @@
+Various improvements to distribution package test support:
+
+* Create three separate implementations for how to access reference metadata:
+
+  * From a file on PyPI
+  * From a wheel
+  * From an sdist
+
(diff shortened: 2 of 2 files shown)


diazona:
The `needs_network` marker should take care of that. Something like `-k 'not needs_network'` passed to pytest, I think.

I doubt that this is going to be packaged by any Linux distributions though. If that winds up happening we can cross the proverbial bridge when we come to it.

PR #127, opened 2023-12-05
Change the primary name of the console script to setuptools-pyproject-migration

### LINKED ISSUE #107: Change name of console script to match name of package
The `pipx run` command makes it very convenient to download a package and run a script entry point which has the same name as the package. If we want to take advantage of that convenience (and I'd say we do), we should change the name of our console script from `setup-to-pyproject` to `setuptools-pyproject-migration`. (Or at least we need to add a console script with that name, but I don't think there's any point in having two scripts that do the same thing, except for backward compatibility but we don't really have to care about that while the package is at such an early stage of development.)

diazona:
This PR changes the name of the console script that users should use to match the name of the package, `setuptools-pyproject-migration`, to support `pipx run`.

We could have kept multiple console script names, but I thought that was more confusion than it's worth.

The intent will be to remove the old console script name in some future version, though I'm not sure if it makes sense to advertise a specific version in which it will be removed or how long we should wait before actually removing it. (Especially given that it didn't even seem to work properly, see #100)

Closes #107

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,12 +43,18 @@ After that, in any directory that has a `setup.py` and/or `setup.cfg` file, you
 can run
 
 ```console
-setup-to-pyproject
+setuptools-pyproject-migration
 ```
 
 and it will print out the content of `pyproject.toml` as computed from your
-`setup.py` and/or `setup.cfg`. Running `setup-to-pyproject -h` will print
-a brief usage summary.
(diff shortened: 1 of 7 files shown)


diazona:
I know we had a bit of discussion about this on the issue so I'm happy to iterate on this change if we could/should be doing it a different way.

sjlongland (approved):

PR #137, opened 2023-12-11
Update tests to use Python 3.12 as the latest stable version

diazona:
This PR includes a couple changes to the Github Actions workflows so that they use Python 3.12 as the latest stable version. When these workflows were written, 3.12 had not been finalized yet so we were using 3.11 as the latest stable version.

PR #140, opened 2023-12-13
Work around incompatibility between Python 3.12 and setuptools<66.1

### LINKED ISSUE #139: "No module named 'distutils'" with old setuptools on Python 3.12
On Python 3.12, when running tests with old versions of setuptools (using `--force-dep` in tox), specifically version 56, the pre-testing setup fails with this error:

```console
$ tox -r -e py312 --force-dep setuptools==56
[...]
py312: commands[0]> pytest tests/test_classifiers.py
ImportError while loading conftest '/home/diazona/programming/setuptools-pyproject-migration/tests/conftest.py'.
tests/conftest.py:4: in <module>
    import test_support
test_support/test_support/__init__.py:1: in <module>
    import distutils.core
E   ModuleNotFoundError: No module named 'distutils'
py312: exit 4 (0.40 seconds) /home/diazona/programming/setuptools-pyproject-migration> pytest tests/test_classifiers.py pid=31540
```

Discovered in [this example in CI](https://github.com/diazona/setuptools-pyproject-migration/actions/runs/7167025692/job/19512272944).

diazona:
Older versions of setuptools fail to import distutils on Python 3.12, presumably due to the removal of distutils from the standard library. This bug was fixed in https://github.com/pypa/setuptools/pull/3685, and setuptools 66.1 was the first version released that includes the fix. (I'm not sure exactly how that PR fixed the bug, but it doesn't matter.)

Because setuptools 66.0 understandably does not declare itself to be incompatible with Python 3.12, I'm bumping our minimum setuptools version requirement on that Python version to reflect the incompatibility, and adjusting the CI configuration to match.

Closes #139

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,6 +80,14 @@ jobs:
         # Note the -- used to separate tox options from pytest options.
         run: >
           tox -e py
-          ${{ inputs.force-minimum-dependencies && '--force-dep setuptools==56 --force-dep setuptools_scm==3.4.1 --force-dep typing-extensions==4' || '' }}
+          ${{
+            inputs.force-minimum-dependencies
+            && (
+              inputs.python-version >= '3.12'
+              && '--force-dep setuptools==66.1 --force-dep setuptools_scm==3.4.1 --force-dep typing-extensions==4'
+              || '--force-dep setuptools==56 --force-dep setuptools_scm==3.4.1 --force-dep typing-extensions==4'
+            )
+            || ''
+          }}
           --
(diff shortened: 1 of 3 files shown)


diazona:
I rebased #137 on top of this and checked that its tests pass. :+1:

sjlongland (approved):
Gotta love it when an upstream change pulls the rug on a downstream project. :-)

diazona:
Heh, indeed... But it happens, and in this case it's really a good change overall IMO, so no complaints.

Issue #145, opened 2023-12-21
ModuleNotFoundError: No module named 'setuptools_scm._integration'


diazona:
I'm seeing an error `ModuleNotFoundError: No module named 'setuptools_scm._integration'` when running the package distribution tests for `aioax25`, but only under very specific circumstances: it only happens with Python 3.12, and only when running all the tests or some large subset of them - if I try to rerun with the specific tests that fail, the error doesn't show up.

```console
$ tox -e py312 --force-dep setuptools==66.1 --force-dep setuptools_scm==3.4.1 --force-dep typing-extensions==4 --
[...]
__________________________________________________________________________________________ ERROR at setup of TestExternalProject.test_name[aioax25==0.0.11.post0] __________________________________________________________________________________________

self = <test_distribution_packages.TestExternalProject object at 0x7f8b21d78530>, distribution_package = <test_support.distribution.PyPiPackagePreparation object at 0x7f8b1ce44380>

    @pytest.fixture(scope="class")
    def actual(self, distribution_package: DistributionPackagePreparation) -> StandardMetadata:
>       return StandardMetadata.from_pyproject(distribution_package.project.generate())

/home/diazona/programming/setuptools-pyproject-migration/tests/distribution/test_distribution_packages.py:91: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/diazona/programming/setuptools-pyproject-migration/test_support/test_support/__init__.py:192: in generate
    distribution: distutils.dist.Distribution = distutils.core.run_setup(
/home/diazona/programming/setuptools-pyproject-migration/.tox/py312/lib/python3.12/site-packages/setuptools/_distutils/core.py:267: in run_setup
    exec(code, g)
<string>:19: in <module>
    ???
/home/diazona/programming/setuptools-pyproject-migration/.tox/py312/lib/python3.12/site-packages/setuptools/__init__.py:108: in setup
[...]

PR #146, opened 2023-12-27
Conditionally disable the pytest distribution package test

diazona:
When running tests under Python 3.12 with setuptools_scm==3.4.1, we run into a problem where a call to `setuptools.setup()` tries to load a newer version of setuptools_scm, and that conflicts with the older version that is already in sys.modules. I haven't fully figured out the details, but the gist is that setuptools attempts to read and install the entry point specifications from the newer version of setuptools_scm while using the older version's code, and there is one particular entry point whose value refers to the setuptools_scm._integration module which doesn't exist in the older version's code. This winds up breaking every test that runs after the older version of setuptools_scm gets loaded.

We have various options for fixing this, but I don't want to hold up the release any longer while figuring out what the best way to do it is. (I'm guessing that the best approach is just to use virtual environments for distribution package tests, but I want to explore the options.) So I'm disabling the distribution package test which triggers this error when using Python>=3.12 and setuptools_scm<6. We can implement a proper fix later.

---

Note that I omitted a changelog fragment because I don't really want to have to regenerate the changelog just for this, and anyway this PR doesn't fix the issue, it just postpones dealing with the problem.

Related to #145 (but does not close it)

### DIFF
--- a/tests/distribution/test_distribution_packages.py
+++ b/tests/distribution/test_distribution_packages.py
@@ -21,6 +21,7 @@ from PyPI. But verifying that the procedure above works is the ultimate goal.
 """
 
 import pytest
+import sys
 
 from typing import Iterator, List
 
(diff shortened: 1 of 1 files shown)


diazona:
I guess we won't really know if this solves the problem until we try making a release again, which is something to fix for the future, but for now probably best to just go ahead with it. I will try making a release candidate before the real release to make sure.

sjlongland (approved):
Yep, if it's going to be problematic, let's disable it for now. :-)

PR #147, opened 2024-01-15
Reduce the pre-commit autoupdate frequency to monthly

### LINKED ISSUE #109: Reduce frequency of pre-commit autoupdates
Right now every week we get a PR from the pre-commit CI service that tries to update all our hooks, including updating the ruff hook to a version which is incompatible with Python 3.6. We probably don't need weekly updates to this configuration, as there are basically never serious problems with the hooks that require an immediate update, so this issue would entail changing it to monthly or quarterly to reduce how often we have to deal with those PRs. (I've been meaning to make that change for a long time but never got around to it, just out of laziness.)

Or we could stop using the pre-commit.ci service entirely and just implement our own pre-commit workflow in this repo. That would let us avoid automatic updates entirely. (In fact we could even write our own separate logic to handle them, although that would be outside the scope of this issue.)

diazona:
It's kind of annoying to deal with weekly updates because we always have to revert the changes to the ruff hook, at least as long as we want to keep supporting Python 3.6. So I'm changing it to monthly. We could decide to further reduce the frequency or drop usage of pre-commit.ci entirely in the future.

Closes #109

I'd probably recommend merging this after #138, but I'm not sure if it matters.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,6 +55,7 @@ repos:
   - id: towncrier-check
     files: $changelog\.d/
 ci:
+  autoupdate_schedule: monthly
   skip:
   # pre-commit.ci doesn't have Docker available
   - actionlint-docker
--- a/changelog.d/109.misc.rst
+++ b/changelog.d/109.misc.rst
@@ -1,0 +1,1 @@
+Reduce pre-commit.ci autoupdate frequency to monthly


sjlongland (approved):

sjlongland:
Sounds sane to me, we don't need to be on the bleeding edge. :-)

PR #153, opened 2024-06-08
Handle INI-style strings for `entry_points`

### LINKED ISSUE #152: AttributeError: 'str' object has no attribute 'items' when running.
I wanted to try this app but I'm getting the above error.
Tried it in a python3.11 environment, as well as a 3.9 one.

Following is the traceback, along with the python3 and pip versions, as well as the output of pip3 list:

```bash
setuptools-pyproject-migration 
running pyproject
  File "/Users/Koji/.local/bin/setuptools-pyproject-migration", line 8, in <module>
    sys.exit(main())
  File "/Users/Koji/.local/pipx/venvs/setuptools-pyproject-migration/lib/python3.9/site-packages/setuptools_pyproject_migration/cli.py", line 67, in main
    exec(setup_bytecode)
  File "setup.py", line 27, in <module>
    setup(
  File "/Users/Koji/.local/pipx/shared/lib/python3.11/site-packages/setuptools/__init__.py", line 87, in setup
    return distutils.core.setup(**attrs)
  File "/Users/Koji/.local/pipx/shared/lib/python3.11/site-packages/setuptools/_distutils/core.py", line 185, in setup
    return run_commands(dist)
  File "/Users/Koji/.local/pipx/shared/lib/python3.11/site-packages/setuptools/_distutils/core.py", line 201, in run_commands
    dist.run_commands()
  File "/Users/Koji/.local/pipx/shared/lib/python3.11/site-packages/setuptools/_distutils/dist.py", line 968, in run_commands
    self.run_command(cmd)
  File "/Users/Koji/.local/pipx/shared/lib/python3.11/site-packages/setuptools/dist.py", line 1217, in run_command
    super().run_command(command)
  File "/Users/Koji/.local/pipx/shared/lib/python3.11/site-packages/setuptools/_distutils/dist.py", line 987, in run_command
    cmd_obj.run()
  File "/Users/Koji/.local/pipx/venvs/setuptools-pyproject-migration/lib/python3.9/site-packages/setuptools_pyproject_migration/__init__.py", line 403, in run
    tomlkit.dump(self._generate(), sys.stdout)
  File "/Users/Koji/.local/pipx/venvs/setuptools-pyproject-migration/lib/python3.9/site-packages/setuptools_pyproject_migration/__init__.py", line 383, in _generate
    entry_points = _generate_entry_points(dist.entry_points)
  File "/Users/Koji/.local/pipx/venvs/setuptools-pyproject-migration/lib/python3.9/site-packages/setuptools_pyproject_migration/__init__.py", line 74, in _generate_entry_points
    for eptype, raweps in entry_points.items():
AttributeError: 'str' object has no attribute 'items'

```

```bash
$ python3 -V
Python 3.11.9
$ pip3 -V
pip 24.0 from /opt/local/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pip (python 3.11)
$ pip3 list
Package                   Version
------------------------- ---------
altgraph                  0.17.4
anytree                   2.12.1
argcomplete               3.3.0
attrs                     23.2.0
Automat                   20.2.0
bcrypt                    4.1.2
build                     1.0.3
calver                    2024.4.10
cffi                      1.16.0
click                     8.1.7
constantly                15.1.0
cryptography              42.0.5
Cython                    3.0.10
docutils                  0.20.1
editables                 0.5
fancycompleter            0.9.1
flit_core                 3.9.0
hatch-fancy-pypi-readme   24.1.0
hatch-vcs                 0.4.0
hatchling                 1.24.2
hyperlink                 21.0.0
idna                      3.6
incremental               22.10.0
installer                 0.7.0
Jinja2                    3.1.3
libtorrent                2.0.10
libxml2-python            2.12.6
lxml                      5.2.1
m2r                       0.2.1
macholib                  1.16.3
Mako                      1.3.2
Markdown                  3.6
markdown-it-py            3.0.0
MarkupSafe                2.1.5
mdurl                     0.1.2
meson                     1.4.0
meson-python              0.15.0
mistune                   3.0.2
modulegraph               0.19.6
modulegraph2              2.3
numpy                     1.26.4
objectgraph               1.0.3
oldest-supported-numpy    0.1
olefile                   0.47
packaging                 24.0
Parsley                   1.3
pathspec                  0.12.1
pdb-attach                3.0.1
pdbpp                     0.10.2
pep508-parser             2019.3
Pillow                    9.5.0
pip                       24.0
platformdirs              4.2.2
pluggy                    1.4.0
poetry-core               1.9.0
py2app                    2.0a0
pyasn1                    0.5.1
pyasn1_modules            0.3.0
pycairo                   1.26.0
pycparser                 2.22
pygame                    2.5.2
Pygments                  2.17.2
PyGObject                 3.46.0
PyHamcrest                2.0.2
pyOpenSSL                 24.1.0
pyproject_hooks           1.1.0
pyproject-metadata        0.7.1
pyrepl                    0.9.0
pyxdg                     0.28
PyYAML                    6.0.1
rencode                   1.0.6
rich                      13.7.1
roman                     4.1
semantic_version          2.10.0
service-identity          23.1.0
setproctitle              1.2.2
setupmeta                 3.6.0
setuptools                69.5.1
setuptools-git-versioning 2.0.0
setuptools-rust           1.8.1
setuptools-scm            8.0.4
six                       1.16.0
smartypants               2.0.1
toml                      0.10.2
tomlkit                   0.12.5
trove-classifiers         2024.4.10
Twisted                   24.3.0
typing_extensions         4.11.0
typogrify                 2.0.7
userpath                  1.9.2
wheel                     0.43.0
wmctrl                    0.3
zope.interface            6.1
```

sjlongland:
A number of Python projects use this somewhat ugly syntax for specifying entry points.  So we need to support it sadly.

Closes #152.

### DIFF
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -4,6 +4,7 @@ import setuptools
 import sys
 import tomlkit
 import warnings
+import configparser
 from packaging.specifiers import SpecifierSet
 from pep508_parser import parser as pep508
 from tomlkit.api import Array, InlineTable
@@ -51,12 +52,22 @@ def _parse_entry_point(entry_point: str) -> Tuple[str, str]:
     return (name.strip(), target.strip())
 
 
-def _generate_entry_points(entry_points: Optional[Dict[str, List[str]]]) -> Dict[str, Dict[str, str]]:
+def _generate_entry_points(entry_points: Optional[Union[Dict[str, List[str]], str]]) -> Dict[str, Dict[str, str]]:
     """
     Dump the entry points given, if any.
 
     >>> _generate_entry_points(None)
     {}
+    >>> _generate_entry_points('''
+    ...     [type1]
+    ...     ep1=mod:fn1
+    ...     ep2=mod:fn2
+    ...
+    ...     [type2]
+    ...     ep3=mod:fn3
+    ...     ep4=mod:fn4
+    ... ''')
+    {'type1': {'ep1': 'mod:fn1', 'ep2': 'mod:fn2'}, 'type2': {'ep3': 'mod:fn3', 'ep4': 'mod:fn4'}}
     >>> _generate_entry_points({"type1": ["ep1=mod:fn1", "ep2=mod:fn2"],
     ...                        "type2": ["ep3=mod:fn3", "ep4=mod:fn4"]})
     {'type1': {'ep1': 'mod:fn1', 'ep2': 'mod:fn2'}, 'type2': {'ep3': 'mod:fn3', 'ep4': 'mod:fn4'}}
@@ -71,8 +82,17 @@ def _generate_entry_points(entry_points: Optional[Dict[str, List[str]]]) -> Dict
 
     parsed_entry_points: Dict[str, Dict[str, str]] = {}
 
-    for eptype, raweps in entry_points.items():
-        parsed_entry_points[eptype] = dict(map(_parse_entry_point, raweps))
+    if isinstance(entry_points, str):
+        # INI-style
+        parser = configparser.ConfigParser()
+        parser.read_string(entry_points)
+        for eptype, section in parser.items():
+            if eptype != "DEFAULT":
+                parsed_entry_points[eptype] = dict(section.items())
+    else:
+        # dict
+        for eptype, raweps in entry_points.items():
+            parsed_entry_points[eptype] = dict(map(_parse_entry_point, raweps))
 
     return parsed_entry_points
 
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -66,11 +66,31 @@ def _generate_entry_points(entry_points: Optional[Union[Dict[str, List[str]], st
     ...     [type2]
     ...     ep3=mod:fn3
     ...     ep4=mod:fn4
-    ... ''')
-    {'type1': {'ep1': 'mod:fn1', 'ep2': 'mod:fn2'}, 'type2': {'ep3': 'mod:fn3', 'ep4': 'mod:fn4'}}
+    ... ''')  # doctest: +NORMALIZE_WHITESPACE
+    {'type1': {'ep1': 'mod:fn1', 'ep2': 'mod:fn2'},
+     'type2': {'ep3': 'mod:fn3', 'ep4': 'mod:fn4'}}
+    >>> _generate_entry_points('''
+    ...     [DEFAULT]
+    ...     ep0=mod:fn0
+    ...
+    ...     [type1]
+    ...     ep1=mod:fn1
+    ...     ep2=mod:fn2
+    ...
+    ...     [type2]
+    ...     ep3=mod:fn3
+    ...     ep4=mod:fn4
+    ... ''')  # doctest: +NORMALIZE_WHITESPACE
+    {'DEFAULT': {'ep0': 'mod:fn0'},
+     'type1': {'ep1': 'mod:fn1', 'ep2': 'mod:fn2'},
+     'type2': {'ep3': 'mod:fn3', 'ep4': 'mod:fn4'}}
     >>> _generate_entry_points({"type1": ["ep1=mod:fn1", "ep2=mod:fn2"],
-    ...                        "type2": ["ep3=mod:fn3", "ep4=mod:fn4"]})
-    {'type1': {'ep1': 'mod:fn1', 'ep2': 'mod:fn2'}, 'type2': {'ep3': 'mod:fn3', 'ep4': 'mod:fn4'}}
+    ...                        "type2": ["ep3=mod:fn3", "ep4=mod:fn4"]})  # doctest: +NORMALIZE_WHITESPACE
+    {'type1': {'ep1': 'mod:fn1', 'ep2': 'mod:fn2'},
+     'type2': {'ep3': 'mod:fn3', 'ep4': 'mod:fn4'}}
+    >>> _generate_entry_points({"type1": ["ep1=mod:fn1", "ep2=mod:fn2"],
+    ...                        "type2": []})
+    {'type1': {'ep1': 'mod:fn1', 'ep2': 'mod:fn2'}}
 
     :param: entry_points The `entry_points` property from the
                         :py:class:setuptools.dist.Distribution being examined.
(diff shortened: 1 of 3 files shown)


src/setuptools_pyproject_migration/__init__.py
```diff
@@ -71,8 +82,17 @@ def _generate_entry_points(entry_points: Optional[Dict[str, List[str]]]) -> Dict
...
     parsed_entry_points: Dict[str, Dict[str, str]] = {}
 
-    for eptype, raweps in entry_points.items():
-        parsed_entry_points[eptype] = dict(map(_parse_entry_point, raweps))
+    if isinstance(entry_points, str):
+        # INI-style
+        parser = configparser.ConfigParser()
+        parser.read_string(entry_points)
+        for eptype, section in parser.items():
+            if eptype != "DEFAULT":
```
diazona, line 90: I wonder what `setuptools` would do if it gets passed an entrypoints specification with a `DEFAULT` section? If it doesn't just crash or fail, that would probably be worth adding a test case for.

diazona (commented):
Makes sense, I just think it'd be good to have some test cases for this added under the `tests/` directory. (Whether in an existing file or a new one, whatever makes more sense.)

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -71,8 +82,17 @@ def _generate_entry_points(entry_points: Optional[Dict[str, List[str]]]) -> Dict
 
     parsed_entry_points: Dict[str, Dict[str, str]] = {}
 
-    for eptype, raweps in entry_points.items():
-        parsed_entry_points[eptype] = dict(map(_parse_entry_point, raweps))
+    if isinstance(entry_points, str):
+        # INI-style
+        parser = configparser.ConfigParser()
+        parser.read_string(entry_points)
```
diazona, line 88: Oh and I just noticed in the documentation that [`pkg_resources.EntryPoint.parse_map()`](https://setuptools.pypa.io/en/latest/pkg_resources.html#creating-and-parsing) is a thing. I'm guessing that's the same logic setuptools uses to parse these string entry point specifications, and if so we could use it. On the other hand it might be a good idea to avoid any dependencies on pkg_resources because it _is_ deprecated :shrug: but given that the whole point of this package is to work with deprecated-ish configuration files, maybe that's okay.

src/setuptools_pyproject_migration/__init__.py
diazona: Since I was doing some tests anyway, I tried it out: it turns out `pkg_resources.EntryPoint.parse_map()` will happily accept a section named `DEFAULT` and create a corresponding entry point group, which probably suggests that that's what setuptools will do as well. I didn't actually create a project to test it, but I think this is definitely worth having a test case for.

sjlongland: Yeah, I assumed that the `DEFAULT` group would be pretty meaningless, but then again, this would not be the first time such assumptions bit us.

I think it's supported by accident rather than by design.

We could remove that `if` block… actually I'd probably replace the condition with a check to see if the resulting group is empty so we don't needlessly create empty groups.

src/setuptools_pyproject_migration/__init__.py
sjlongland: Yeah, I did see mention of a `split_sections()` "utility" function, but so far haven't uncovered it.  Otherwise that's exactly what I would have used (or at least seen what its implementation was).

I just went digging for it now…
https://github.com/pypa/setuptools/blob/main/pkg_resources/__init__.py#L3361-L3384

Looks like it'd return `(None, [lines_of_text])` in the case of a missing section name, not `DEFAULT` like `configparser` does.

When I look at how it's used in `pkg_resources`, it looks like this is not a supported configuration:

https://github.com/pypa/setuptools/blob/main/pkg_resources/__init__.py#L2686-L2700

While we've got logic to parse out the individual lines that in theory duplicates what's being done here, it's code that's in the standard Python library.

In fact, one approach would be to use `configparser` to re-implement `_parse_entry_point`: use `"\n".join(raweps)` to bundle everything up and let `configparser` extract out the item values.

src/setuptools_pyproject_migration/__init__.py
diazona: Yeah checking for empty groups sounds more practical.

src/setuptools_pyproject_migration/__init__.py
diazona: When I tested, both `EntryPoint.parse_map()` and `ConfigParser.read_string()` raise a `ValueError` if the section name is completely missing, so I'm fine with not really worrying about how to handle that case. I mean, presumably people are running this on valid setuptools configurations which have been used to build and distribute real projects, so it kind of doesn't matter how we handle invalid configurations (as long as it's not, like, deleting the user's home directory :laughing:). At this point I'm thinking more about how to handle an explicitly written section which is actually called `DEFAULT`.

Reimplementing the whole thing with `ConfigParser` is interesting to consider, but personally I wouldn't... this is definitely an opinion, but I'd rather avoid "unparsing" and then parsing an object unless there's a really significant benefit we get from it, and I don't think this meets that threshold.

sjlongland: Fair point … I can roll back that commit. :-)

sjlongland: Actually, I note the `DEFAULT` section has special behaviour… when included, it populates all other sections with its content.

So if I do:

```
[DEFAULT]
ep0=mod:fn0

[type1]
ep1=mod:fn1
ep2=mod:fn2

[type2]
ep3=mod:fn3
ep4=mod:fn4
```

That gets passed through as:

```
[DEFAULT]
ep0=mod:fn0

[type1]
ep1=mod:fn1
ep2=mod:fn2
ep0=mod:fn0

[type2]
ep3=mod:fn3
ep4=mod:fn4
ep0=mod:fn0
```

To leave things as they are and treat `DEFAULT` like any other section, one must pass `default_section=None` in the constructor of `ConfigParser`.  Otherwise you get this "special" treatment.

https://stackoverflow.com/a/71891193

So if I do this, I can have a section named `DEFAULT`; and furthermore, if I strip "empty" groups, the result will behave the way we expect.

sjlongland: `default_section=""` works, and keeps `mypy` happy.

diazona (approved):
Works for me. I would kind of like to have more tests but I can always add those later.

PR #155, opened 2024-07-03
Update the project status to active

diazona:
This should have been done when we released our first version but we forgot, I guess.


sjlongland (approved):
Yep, looks good to go.

Issue #132, opened 2023-12-10
KeyError: 'long_description'


diazona:
When running the distribution package test for `pytest-localserver`, we get a `KeyError` on line 311 of `__init__.py`:

```python
                long_description_source = dist.command_options["metadata"]["long_description"][1]
```

To reproduce, run

```console
$ tox run -e py311 -- 'tests/distribution/test_distribution_packages.py' -k pytest-localserver --runxfail
```

(or same on any other Python version)

PR #156, opened 2024-07-07
Add more robust handling of long descriptions

diazona:
This PR makes some major improvements in how we handle the `long_description` field from setuptools:

- Now if the long description is given as an inline string in `setup.cfg` or `setup.py` (as far as we can tell) rather than being loaded from a file, it will be kept as an inline string in `pyproject.toml`.
- If there's no content type provided and one cannot be definitively determined from the file extension, the user can provide it with a new option `--readme-content-type`; otherwise the program will guess `text/plain`.
- It fixes the error where the long description would sometimes be missing from the command options, closing #132.
- The code is refactored to be more modular and more testable.

Thanks to contributors on [a Mastodon thread](https://techhub.social/@diazona/112736102428118730) for some useful discussion.

### DIFF
--- a/src/setuptools_pyproject_migration/__init__.py
+++ b/src/setuptools_pyproject_migration/__init__.py
@@ -7,21 +7,9 @@ import tomlkit
 import warnings
 from packaging.specifiers import SpecifierSet
 from pep508_parser import parser as pep508
+from setuptools_pyproject_migration._types import Contributor, Pyproject
 from tomlkit.api import Array, InlineTable
-from typing import Dict, List, Optional, Set, Tuple, Type, TypeVar, Union
-
-# After we drop support for Python <3.10, we can import TypeAlias directly from typing
-from typing_extensions import Required, TypedDict
-
-
-# PEP 518
-BuildSystem: Type = TypedDict("BuildSystem", {"requires": List[str], "build-backend": str}, total=True)
-
-# https://packaging.python.org/en/latest/specifications/declaring-project-metadata/
-Contributor: Type = TypedDict("Contributor", {"name": str, "email": str}, total=False)
-LicenseFile: Type = TypedDict("LicenseFile", {"file": str})
-LicenseText: Type = TypedDict("LicenseText", {"text": str})
-ReadmeInfo: Type = TypedDict("ReadmeInfo", {"file": str, "content-type": str})
+from typing import Dict, List, Optional, Set, Tuple, TypeVar, Union
 
 
 def _parse_entry_point(entry_point: str) -> Tuple[str, str]:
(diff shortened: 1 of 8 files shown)


sjlongland:
@diazona No problems, well I can review and hold off on actual merging if you think there might be last-minute changes. :-)

src/setuptools_pyproject_migration/__init__.py
```diff
@@ -171,15 +128,79 @@ def _tomlkit_inlinify(value: T) -> Union[T, Array, InlineTable]:
...
+    True
+    >>> _looks_like_media_type("text/x-rst")
+    True
+    >>> _looks_like_media_type("text/plain")
+    True
+
+    This doesn't actually check whether the type is a real type that's registered
+    with IANA or in common use (use the :py:module:`mimetypes` module for that),
+    only whether it follows the syntax of a media type. So even unknown types
+    will return true.
```
sjlongland, line 167: Probably safest doing this, people can and do invent their own weird and wonderful MIME types, as the system is extensible.

src/setuptools_pyproject_migration/_long_description.py
```diff
@@ -0,0 +1,260 @@
...
+        # filename this way because setup() might never receive the
+        # filename if the file was read directly by setup.py.
+        return dist.get_long_description()
+
+
+def _read_long_description_file(path: pathlib.Path) -> str:
+    """
+    Read the content of the long description from a filename.
+    """
+    # TODO handle multiple comma-separated filenames
```
sjlongland, line 145: I'll just put an issue here so we don't forget about this later… no rush of course.

https://github.com/diazona/setuptools-pyproject-migration/issues/158

sjlongland (approved):
Looks good, but I'll hold off on the merge until you're happy with it all as well. :-)

src/setuptools_pyproject_migration/__init__.py
diazona: Yeah, although custom types are supposed to follow specific formats (with `x.` or `x-` or similar).

I created #159 for adding a warning if the script detects a nonstandard content type, as a reminder to consider that.

diazona:
Ah you know what, it looks like I goofed up the commit organization (checked with `git rebase --exec 'pre-commit run --all-files' main`). I'll fix that and re-push - the overall diff will be the same, I just have to move a few patch fragments from one commit to another.

diazona:
OK done! I can go ahead and merge this, since you already approved. (Which brings to mind the separate issue of whether approvals should be reset when the branch is rebased; I feel like it's more "technically correct" to do so, but it really doesn't matter in this context.)

PR #162, opened 2024-07-21
Add explicit coverage.py configuration

diazona:
This pull request adds a coverage.py configuration which enables branch coverage and limits coverage reporting to files in the package itself. Previously, there was nothing telling coverage.py not to measure coverage for files generated during testing, so it would pollute the output with meaningless coverage numbers for a bunch of temporary files.
    
I put the configuration in `pyproject.toml` rather than any of the other configuration files coverage.py supports because TOML is a better specified format than INI.

I also edited the `pytest` command run from tox to ensure that it will use `pyproject.toml` as the coverage.py configuration file.

### DIFF
--- a/changelog.d/+coverage-config.misc.rst
+++ b/changelog.d/+coverage-config.misc.rst
@@ -1,0 +1,1 @@
+Add a coverage.py configuration to select which files are measured and enable branch coverage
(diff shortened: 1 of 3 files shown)


sjlongland:
Makes sense to be explicit on these things, there are fewer surprises that way. :-)

PR #163, opened 2024-07-21
Enable tox output in CI

diazona:
It looks like setting `TOX_PARALLEL_NO_SPINNER` to 1 completely disables all output from tox and the commands it runs, except for the final message that an environment ran successfully. So I'm removing that variable from our CI configuration. We want the full output to appear in CI.


sjlongland:
Yeah, this is a regular pet peeve I have with people at work… "lets hide all the log messages because they're annoying" … something breaks in CI … "Why won't my test pass?!?!"

sjlongland (approved):

PR #168, opened 2024-07-29
Relax the check of optional dependencies to allow missing extras

### LINKED ISSUE #134: Optional dependencies not listed with their corresponding extras
When running the test `tests/distribution/test_distribution_packages.py::TestExternalProject::test_optional_dependencies[pytest==7.3.0]`, setuptools-pyproject-migration does not list the corresponding extras as part of the `Requirement` objects.

I'm not sure what the correct approach should be; we can figure that out while solving this issue.

```
self = <test_distribution_packages.TestExternalProject object at 0x7f32b0ca7090>
expected = StandardMetadata(name='pytest', version=<Version('7.3.0')>, description='pytest: simple powerful testing with Python',...are Development :: Testing', 'Topic :: Utilities'], keywords=['test,unittest'], scripts={}, gui_scripts={}, dynamic=[])
actual = StandardMetadata(name='pytest', version=<Version('7.3.0')>, description='pytest: simple powerful testing with Python',..., 'unittest'], scripts={'pytest': 'pytest:console_main', 'py.test': 'pytest:console_main'}, gui_scripts={}, dynamic=[])

    def test_optional_dependencies(self, expected: StandardMetadata, actual: StandardMetadata):
        # Check that the set of extras is the same
        assert expected.optional_dependencies.keys() == actual.optional_dependencies.keys()
    
        # Check that the dependencies associated with each extra are the same
        for extra in expected.optional_dependencies:
>           assert sorted(expected.optional_dependencies[extra], key=str) == sorted(
                actual.optional_dependencies[extra], key=str
            )
E           assert [<Requirement('argcomplete; extra == "testing"')>, <Requirement('attrs>=19.2.0; extra == "testing"')>, <Requirement('hypothesis>=3.56; extra == "testing"')>, <Requirement('mock; extra == "testing"')>, <Requirement('nose; extra == "testing"')>, <Requirement('pygments>=2.7.2; extra == "testing"')>, <Requirement('requests; extra == "testing"')>, <Requirement('xmlschema; extra == "testing"')>] == [<Requirement('argcomplete')>, <Requirement('attrs>=19.2.0')>, <Requirement('hypothesis>=3.56')>, <Requirement('mock')>, <Requirement('nose')>, <Requirement('pygments>=2.7.2')>, <Requirement('requests')>, <Requirement('xmlschema')>]
E             At index 0 diff: <Requirement('argcomplete; extra == "testing"')> != <Requirement('argcomplete')>
E             Full diff:
E               [
E             -  <Requirement('argcomplete')>,
E             +  <Requirement('argcomplete; extra == "testing"')>,
E             ?                           ++++++++++++++++++++
E             -  <Requirement('attrs>=19.2.0')>,
E             +  <Requirement('attrs>=19.2.0; extra == "testing"')>,
E             ?                             ++++++++++++++++++++
E             -  <Requirement('hypothesis>=3.56')>,
E             +  <Requirement('hypothesis>=3.56; extra == "testing"')>,
E             ?                                ++++++++++++++++++++
E             -  <Requirement('mock')>,
E             -  <Requirement('nose')>,
E             +  <Requirement('mock; extra == "testing"')>,
E             +  <Requirement('nose; extra == "testing"')>,
E             -  <Requirement('pygments>=2.7.2')>,
E             +  <Requirement('pygments>=2.7.2; extra == "testing"')>,
E             ?                               ++++++++++++++++++++
E             -  <Requirement('requests')>,
E             -  <Requirement('xmlschema')>,
E             +  <Requirement('requests; extra == "testing"')>,
E             +  <Requirement('xmlschema; extra == "testing"')>,
E               ]

/home/diazona/programming/setuptools-pyproject-migration/tests/distribution/test_distribution_packages.py:146: AssertionError
```

diazona:
We've had failures in the distribution package tests because the expected optional dependencies (parsed from the packages' core metadata by our test support code) include markers for the extras, but the actual optional dependencies (generated by `setuptools_pyproject_migration` itself) don't. This commit addresses those failures by effectively trying to ignore the part of the marker expression that specifies an extra and comparing the rest.

The most robust way to fix this would have been to add some code that parses the markers into a tree and manipulates the tree to add or remove the extra marker. But parsing markers is pretty complicated; the `packaging` package includes some code to do it, but it takes up two [whole](https://github.com/pypa/packaging/blob/main/src/packaging/_parser.py) [modules](https://github.com/pypa/packaging/blob/main/src/packaging/_tokenizer.py), and it's not part of their public API so we shouldn't really use it. And we'd still need to do some tree manipulation even after parsing, which would probably involve adding some reasonably complex code to the test support package.
                                                                                                                                                                                                                                                     
Instead, what I've done here is to put some heuristics in the test for optional dependencies which will recognize common patterns by which the expected and actual markers might differ. Specifically, any of these will not cause the test to fail:

- The expected and actual markers are equal (this was the status quo)
- The actual marker is unset and the expected marker only has one condition, which is the extra
- The expected marker is "{actual marker} and extra == {extra}", with parenthesis if needed to enforce precedence; this is nearly the same [pattern used by pyproject-metadata to convert its optional dependency Requirements into RFC822 headers](https://github.com/pypa/pyproject-metadata/blob/main/pyproject_metadata/__init__.py#L372-L382) (although the version I added is not entirely careful about checking for nested parentheses in the actual marker, so it could get confused, but we can fix that if it comes up)

With these heuristics, we can get our optional dependency tests to pass, which as far as I'm concerned is good enough to go forward with.

Closes #134

### DIFF
--- a/changelog.d/134.bugfix.rst
+++ b/changelog.d/134.bugfix.rst
@@ -1,0 +1,1 @@
+Relax matching of optional dependency requirements to allow for a missing extra condition in the marker
--- a/tests/distribution/test_distribution_packages.py
+++ b/tests/distribution/test_distribution_packages.py
@@ -20,6 +20,9 @@ having to build all those wheel files, such as using already-published wheels
 from PyPI. But verifying that the procedure above works is the ultimate goal.
 """
 
+import logging
+import packaging.markers
+import packaging.requirements
 import pytest
 import sys
 
@@ -36,6 +39,9 @@ from test_support.distribution import (  # noqa: E402
     DistributionPackagePreparation,
     PyPiDistribution,
 )
+
+
+_logger = logging.getLogger("setuptools_pyproject_migration:tests:" + __name__)
 
 
 def _setuptools_scm_version_conflict() -> bool:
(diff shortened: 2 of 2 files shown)


sjlongland (approved):
Geez what a mess… anyway, good spot finding out a work-around. :-)

PR #171, opened 2024-08-06
Remove typo in README.

tmzhuang:


sjlongland (approved):
Ahh oops, well spotted. :-)

diazona:
Yep thanks!

One note is that the summary line of the commit message shouldn't generally be written with trailing punctuation, but that's not a big enough deal to block this.

tmzhuang:
> Yep thanks!
> 
> One note is that the summary line of the commit message shouldn't generally be written with trailing punctuation, but that's not a big enough deal to block this. 

Thanks for the note! Will keep it in mind for next time :)

PR #174, opened 2024-08-26
Exclude Python 3.6 on Mac OS

diazona:
GitHub Actions no longer supports Python 3.6 on any version of Mac OS, so we can exclude that from the matrix of tests.

This fixes one of the job failures that came up while testing v0.3.


sjlongland (approved):
Yep, it'd have to be a _very_ old MacOS X release to require Python 3.6… (I think I have 3.8 on my MacBook running OS X 10.6.)

PR #177, opened 2024-08-28
Configure pytest to use importlib import mode

diazona:
pytest recommends using the importlib import mode for new projects to avoid test name conflicts, as described [in the documentation](https://docs.pytest.org/en/latest/explanation/goodpractices.html#which-import-mode). Our test files all have different names, so this hasn't been an issue, but it's still a good idea to use the recommended practice. So I'm changing the pytest configuration to do so.


sjlongland (approved):
Yep, makes sense… head off potential trouble before it happens.

PR #181, opened 2024-08-31
Generate changelog for version 0.3

diazona:
This was done by running `tox -e towncrier -- --version 0.3`.

I've generated the changelog for version 0.3 even though this commit is based on v0.3.1 because the changes made between those two versions didn't have a changelog fragment anyway (they should, but I forgot), and ensuring eventual consistency of the changelog is more important than getting the changelog entries in the right release. Ultimately I just don't think it's important enough to make it worth doing the technically proper thing of making post-releases for both v0.3 and v0.3.1.

My plan is to release v0.3.1.post0 based on this branch _without_ merging in the changes made on `main` after the release of v0.3.1. That way the packages we push to PyPI will look approximately right: the only difference between v0.3.1 and v0.3.1.post0 will be the changelog update.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,27 @@
+0.3
+===
+
+Bugfixes
+--------
+
+- Add more robust handling of long descriptions and their content types (#132)
+- Fix a bug in test support code where it wouldn't parse dependencies without markers from a package's core metadata (#133)
+- Relax matching of optional dependency requirements to allow for a missing extra condition in the marker (#134)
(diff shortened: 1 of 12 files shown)


diazona:
@sjlongland by way of explanation: I'd appreciate your sanity check on this, then if all is in order I'll create the post release (so I can use this branch as the target of the release tag), and _then_ I'll merge the PR and delete the branch.

sjlongland (approved):
Yep, looks like the change-log rendered correctly.

PR #183, opened 2024-09-14
Fix parsing of wheel version numbers

diazona:
In our test support code we have a function that parses sdist and wheel filenames, but it turns out that function had a bug that caused it to miss the number in the prerelease component of a wheel filename. To fix that, I replaced the wheel filename parser with the [wheel-filename](https://github.com/wheelodex/wheel-filename) package, and reduced our custom code to just parsing sdist filenames, which is much easier. I also added tests of the new parsing function.

I discovered this while adding the [extrainterpreters](https://github.com/jsbueno/extrainterpreters) package as a distribution package test case.


diazona:
It took a while to figure out, but I think this one is ready now.

sjlongland (approved):

PR #186, opened 2024-10-12
[DO NOT MERGE] Testing metadata and precommit changes to see if we can get precommit passing.

sjlongland:
I haven't figured out how to trigger this from the Actions directly.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -42,7 +42,9 @@ jobs:
       - uses: actions/checkout@v3
       - name: Setup Python
         uses: actions/setup-python@v4
+      - name: Set up venv
+        run: python -m venv .pyenv
       - name: Install tox
-        run: python -m pip install tox
-      - name: Run
-        run: tox -e docs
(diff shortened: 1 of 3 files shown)


diazona:
FWIW to trigger a GitHub action directly it needs to have `workflow_dispatch` in the list of trigger events... and I think that change may need to be merged into `main` before GitHub will actually recognize it and show the UI element that allows you to run the workflow. (But once that's in place you can run the workflow on any branch)

It probably wouldn't be a bad idea to add `workflow_dispatch` to the list of triggers and merge that, so we have the option to run workflows on arbitrary branches for cases like this. Though I don't know how permissions work in that case - if any random person could just pop in and trigger a workflow run, that wouldn't be great. Ideally it'd be restricted to collaborators.

Also as far as I'm concerned, a PR being in draft status is sufficient to indicate that it shouldn't be merged - and that way GitHub will actually enforce it too. Putting "DO NOT MERGE" in the title, or in a label, doesn't hurt anything but is easy to ignore by accident.

PR #195, opened 2025-07-05
Allow manual release testing

diazona:
In this commit I'm adding a workflow_dispatch trigger to the release workflow, which means at any time we can manually invoke the workflow to run a full suite of pre-release tests. I added conditions to the actual publish jobs so that manually invoked workflows will not attempt to upload to PyPI or Test PyPI, but they will do everything else. I also added the run-name entry in the YAML file to put a more descriptive name on manually triggered runs, so we don't get confused about which runs are actually publishing packages.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,16 @@
 name: Publish package to PyPI
+run-name: >
+  ${{
+    !(github.event_name == 'release' && github.event.action == 'published')
+    && 'Manual pre-release test'
+    || ''
+  }}
 
 on:
   release:
     types:
     - published
+  workflow_dispatch:
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


diazona:
I want to try testing this with [act](https://github.com/nektos/act) to see if it works the way I expect before merging it, but otherwise I don't anticipate any further changes.

sjlongland (approved):
Sounds fair, we don't want pre-releases being accidentally published to `pypi` just because we wanted to run the full integration test suite.

diazona:
I _think_ it works, but act isn't the best about explicitly showing when things are skipped so I'm not sure.

In any case, there seem to be some niche test failures introduced by #194 so it wouldn't even get to the uploading step yet. I'll work on that.

PR #202, opened 2025-09-06
Allow pre-commit hook autoupdate PRs to bypass changelog requirement

diazona:
We periodically have the pre-commit GitHub app make a pull request that updates the pre-commit hooks to their latest versions. Since these PRs are generated completely automatically by an external service, they don't have changelog fragments, and it would be rather silly to make someone add changelog fragments (or build our own companion app to add changelog fragments) to them. So we can exempt them from the requirement for pull requests to have changelog fragments.

In this commit, I'm implementing that exemption. I used a heuristic to detect the automatic pre-commit update PRs: `.pre-commit-config.yaml` has to be the only file modified, and at least one commit in the PR needs to be authored by the bot. This captures both fully automatic PRs as well as PRs where we needed to make a manual commit to get tests to pass. Obviously it's not foolproof; someone could definitely add a commit to an auto-update PR that modifies `.pre-commit-config.yaml` in a way that should have a changelog fragment and get away with not adding one, but you'd have to be trying to do that, and the goal here is only to prevent accidentally forgetting changelog fragments.


diazona:
Silly me, it passed because there's a changelog file in this PR :facepalm: 

I ran some tests locally with `act` and this seems to do what it's supposed to do, so I think it's ready after all.

sjlongland (approved):

PR #210, opened 2026-01-07
Require wheel-filename<2 to avoid import errors

diazona:
In our tests we use the wheel-filename package, which recently released its major version 2, and that version includes API changes that break our test code. To prevent our tests from failing, I'm restricting wheel-filename to versions prior to 2.0 in this commit. A future commit will make the necessary changes for this package to work with wheel-filename 2. (The only reason I'm not doing that now is laziness, really :stuck_out_tongue: I want to make the quickest change that gets the code working again.)


sjlongland (approved):
Yep, as an interim workaround for now.  Looks good.

PR #223, opened 2026-07-22
Remove dependabot configuration

diazona:
Dependabot has made a bunch of pull requests that constrain dependencies to extremely recent versions, but since those dependencies aren't actually built in to our program, we're not getting any security benefit from the stricter constraints. They just make it harder for consumers to use the package.

Since Dependabot isn't doing anything useful for us, I'm removing the configuration file to disable it.


sjlongland (approved):
Yep, making work for no good reason.  Let's ditch it.

